PR #2, opened 2024-02-23
Copy CONTRIBUTING.md from unicode-org template

markusicu:
Copy from https://raw.githubusercontent.com/unicode-org/.github/main/.github/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+# Contributing to this project
+
+## Contributor License Agreement
+
+In order to contribute to this project, the Unicode Consortium must have on file a Contributor License Agreement (CLA) covering your contributions, either an individual or a corporate CLA. Pull Requests, issues, and other contributions will not be merged/accepted until the correct CLA is signed. Which version needs to be signed depends on who owns the contribution being made: you as the individual making the contribution or your employer. **It is your responsibility to determine whether your contribution is owned by your employer.** Please review the [Unicode Intellectual Property, Licensing, & Technical Contribution Policy][policies] for further guidance on which CLA to sign, as well as other information and guidelines regarding the Consortium’s licensing and technical contribution policies and procedures.
+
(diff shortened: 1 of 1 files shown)


markusicu:
> Or do we need to do more set up work to do so (like CLA requirement)?

Well, I just made it public, but when I go back to the branch protections, there are no "status checks" to select.
@srl295 how do I set up the CLA assistant "status check"?

PS @nciric: It's much nicer to reply to GitHub comments from the GitHub UI. Replying by email is messy.

srl295:
> > LGTM, **however** this is only necessary if you're planning to _add to_ CONTRIBUTING.md, though, because by default you already get the inherited version (the URL is in the README and it is presented to users automatically when they open a PR).
> > If you're not planning on making any additional sections to this file, then there's no reason to copy it here.
> 
> Ok, I just assumed that a public repo should have a local copy of that file, so before making this repo public (as was approved) I was looking for what might still need to be done.

Mostly just setting up the README and such.

> I will revert this then.

or close it?

> I haven't tried yet whether I have the power to make this repo public. I guess I will find out :-)

probably not, if not just let me know.

> Do we have instructions for how to set up a new public repo? For example, how to set up the CLA assistant? The “... Internal Policies & Processes ...” doc says “Github does not make the CLA Assistant available in private repos”. Does it get auto-added when making the repo public?

The only instructions are to update the README which I've done.

CLA assistant is already setup organization wide, there's nothing that needs to be done there, other than branch protection to require PRs to pass the cla-assistant check.  But i'd come knocking on the door here if that wasn't in place.

markusicu:
> > I haven't tried yet whether I have the power to make this repo public. I guess I will find out :-)
> 
> probably not, if not just let me know.

I did make it public.

> CLA assistant is already setup organization wide, there's nothing that needs to be done there, other than branch protection to require PRs to pass the cla-assistant check. But i'd come knocking on the door here if that wasn't in place.

I went back to the `main` branch protections and tried to turn on “Require status checks to pass before merging” but got
- No status checks found
- Sorry, we couldn’t find any status checks in the last week for this repository.

Is that expected because we do this for the whole org, not for each repo?

PR #27, opened 2024-04-03
Initial scope document

nciric:
Just enumerating what we decided in the last meeting. Having a document for a stage 1 should help us organize.

Please add comments.

### DIFF
--- a/documents/scope.md
+++ b/documents/scope.md
@@ -6,12 +6,26 @@ Estimated completion 2024.
 
 Based on discussion from [2024-04-02](https://docs.google.com/document/d/1a-777RetR4k2gZKHJ5pG3apWKt86jV7-vU-QChX7lv4/edit?usp=sharing) our stage 1 plan is:
 
-1. Focus on a small set of nouns, sourced from CLDR, e.g. month names, units etc to facilitate faster prototyping.
+1. Focus on a small set of nouns, sourced from CLDR, e.g. month names, units etc to facilitate faster prototyping. See "Noun cases" section for details.
 2. Gender and plural inflection can be handled by current message formatting logic, we can tackle them in later stages.
 3. Adjectives are rarely used in UI strings, we can tackle them in later stages.
 4. Wikidata can be pulled in parallel. We need to resolve licensing with Unicode legal council before committing any data to the repository.
 5. We need to come up with a good lexicon format that is compact and fast to use.
 6. Alignment between sentence parts, e.g. pronouns and other parts, is not in scope.
+
+### Noun cases
+
+Properly casing a noun includes handling its plural form. It also requires additional data, like gender, is noun animate/inanimate etc. We need to provide that data in the lexicon.
+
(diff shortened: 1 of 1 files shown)


documents/scope.md
```diff
@@ -0,0 +1,20 @@
+# Scope and planning
+
+## Stage 1
+
+Estimated completion 2024.
+
+Based on discussion from [2024-04-02](https://docs.google.com/document/d/1a-777RetR4k2gZKHJ5pG3apWKt86jV7-vU-QChX7lv4/edit?usp=sharing) our stage 1 plan is:
+
+1. Focus on a small set of nouns, sourced from CLDR, e.g. month names, units etc to facilitate faster prototyping.
```
grhoten, line 9: I just realized something. There are a lot of languages. Some are more complicated than others. Perhaps start the focus on some languages with comprehensive coverage level in CLDR? I'm fine with starting the focus on English and Spanish, but perhaps some other languages can be considered depending on available resources/knowledge in the group.

documents/scope.md
```diff
@@ -0,0 +1,20 @@
...
+2. Gender and plural inflection can be handled by current message formatting logic, we can tackle them in later stages.
```
grhoten, line 10: Can you clarify what is stated here?

The plural rules are already handled by CLDR plural rules, but inflecting a word to the plural form seems to be a big reason for this group. That's not needed for month names, but inflecting a quantity requires the plural form. Inflecting to the plural form is required for handling units (mentioned above).  Handling units also requires handling grammatical case.

I'm fine with deferring the ability to inflect the word "actor" to "actress" and vis versa, but I'd like to at least allow annotation of the gender of a given word, which is required for articles and prepositions.

documents/scope.md
nciric: I added a new section and listed some potential candidates.

documents/scope.md
nciric: Added a section to clarify. Properly casing a noun includes plural form generation, also at least in Serbian you need gender and additional info to be able to case the noun - so that info would be part of the lexicon.

richgillam: > That's not needed for month names...

Maybe not for month names, but for day names, I could see things like "Three Thursdays from now" being something somebody would want to produce...

richgillam:
FWIW, I think this scope document is a good start, but I'm still a little unclear as to just what our deliverable is here.  Code to do the inflection?  How would that code interact with ICU (if at all) or with MF2?  Or are we just maintaining lexicon data?  Or both?

nciric:
> FWIW, I think this scope document is a good start, but I'm still a little unclear as to just what our deliverable is here. Code to do the inflection? How would that code interact with ICU (if at all) or with MF2? Or are we just maintaining lexicon data? Or both?

I added a section about deliverables. We can update after the ICU meeting tomorrow where we can try to answer some of the MF2.0 questions (e.g. I would say MF2.0 allows for custom formaters, and this can be one of them).

PR #28, opened 2024-04-05
CLDR-json data extraction script

nciric:
Created a script that pulls data from cldr-json (month names and day names) for sr, es, fr, de, en languages.

I appeneded grammatical info, which is correct for sr, but it's probably wrong in day names for other languages. Also, some grammatical information may not be needed for some languages, e.g. I am not sure if animacy matters in English, or are there genders for days/months?

We can land it as is, then file an issue to fix by others.


JelenaMitrovic:
I checked German, English, and Serbian. German looks fine to me - easy cause all month and day names are masculine.

For English, we do not need gender I would say, but we might keep it as a  placeholder, as natural gender might still be needed for some applications. 

Serbian looks good. Are we using Cyrillic only, or should we also include Latin script as they are used more or less equally? I am not sure if Cyrillic only is used in localisation scenarios.

nciric:
Thanks! Serbian Cyrillic is enough as we can generate Latin using ICU
transliterator.

data/de/lexicon.txt
```diff
@@ -0,0 +1,19 @@
+Januar;N;MASC;NOM;SG;INAN
```
grhoten, line 1: Is there a way to use some other format? This is the kind of format that I typically use.
```
Januar: singular accusative dative nominative masculine noun
Januare: plural accusative genitive nominative masculine noun
Januaren: plural dative masculine noun
Januars: singular genitive masculine noun
```
I omitted some mapping details between these entries, but that's mostly how I store this type of data. I also use an XML format that I convert to the syntax above. The XML structure is inverted to the style that I listed above.

Some other formats have been discussed in this group too.

As you expand to languages with numerous grammatical cases, it becomes hard to keep track of how these shouting abbreviations in uppercase map to a grammeme that has to be pronounced in a meeting or for a new person to interact with the data.

The nice feature about using XML is that it's easier to filter and annotate the data as necessary. This becomes important for languages that use bounded morphemes a lot (no whitespace separation) and use different values for the grammatical number for each morpheme.  For example, "your book" in Arabic is a single word with 2 morphemes.  The "your" can be singular, dual or plural. The "your" can be masculine or feminine for pronunciation, depending on who you're addressing. That possessive pronoun binds to the noun being possessed. The word "book" has a gender too, and it can be turned into singular, dual or plural.

nciric: I was going for a simple format that's easy to prototype. Should we use XML from the start? We would need to agree on the basic structure. I'll change the structure now and then we can iterate.

nciric: > Januar: singular accusative dative nominative masculine noun
> Januare: plural accusative genitive nominative masculine noun
> Januaren: plural dative masculine noun
> Januars: singular genitive masculine noun

Changed format to this for now. PTAL.

grhoten: Looks great to me! Thanks.

nciric: Could you approve the PR?

macchiati (approved):

data/tools/extract_cldr_data.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    MONTH_NAMES_EXPRESSION = parse('main..dates.calendars.gregorian.months.format.wide.*')
+    DAY_NAMES_EXPRESSION = parse('main..dates.calendars.gregorian.days.format.wide.*')
+    EXPRESSIONS = [MONTH_NAMES_EXPRESSION, DAY_NAMES_EXPRESSION]
+
+    results = []
+    for expression in EXPRESSIONS:
+        match = expression.find(json_data)
+        for m in match:
+            results.append(m.value + ': noun masculine nominative singular inanimate\n')
```
richgillam, line 60: This doesn't match what I see in the files above.  Are you hand-editing them after the fact, or am I misunderstanding something?

nciric: I generate files with all the grammatical forms, but then delete unnecessary information (depending on the language). So there's no bug - we can expand the tool to know more about each language in the future, and reduce manual work. I leave that to the reader :).

richgillam: Ah, so you _do_ hand-edit.  I wasn't objecting; I was just worried I'd missed something.  :-)

Kutzaki:
In the data for Spanish, these day names are both singular & plural:
lunes;N;MASC;NOM;SG;INAN
martes;N;MASC;NOM;SG;INAN
miércoles;N;MASC;NOM;SG;INAN
jueves;N;MASC;NOM;SG;INAN
viernes;N;MASC;NOM;SG;INAN

data/es/lexicon.txt
```diff
@@ -0,0 +1,19 @@
...
+mayo;N;MASC;NOM;SG;INAN
+junio;N;MASC;NOM;SG;INAN
+julio;N;MASC;NOM;SG;INAN
+agosto;N;MASC;NOM;SG;INAN
+septiembre;N;MASC;NOM;SG;INAN
+octubre;N;MASC;NOM;SG;INAN
+noviembre;N;MASC;NOM;SG;INAN
+diciembre;N;MASC;NOM;SG;INAN
+domingo;N;MASC;NOM;SG;INAN
+lunes;N;MASC;NOM;SG;INAN
```
Kutzaki: Add PL to lunes, martes, miércoles, jueves and viernes

PR #30, opened 2024-04-19
English plural rules using Pynini

nciric:
Implemented plural rules for English using Pynini package to showcase the logic and how-to.

### DIFF
--- a/fst/en/inflection.py
+++ b/fst/en/inflection.py
@@ -4,6 +4,7 @@ Inflects English nouns - creates a plural form from singular.
 Install Pynini package before running.
 """
 
+import os
 import pynini as p
 
 from pynini.lib import pynutil
@@ -18,71 +19,26 @@ _c = p.union('b', 'c', 'd', 'f', 'g', 'h', 'j', 'k', 'l', 'm', 'n',
              'p', 'q', 'r', 's', 't', 'v', 'w', 'x', 'y', 'z')
 _sigma = p.union(_v, _c).closure().optimize()
 
-_exceptions = p.string_map([
-    # Zero plurals.
-    'deer',
-    'fish',
-    'sheep',
-    'series',
-    'species',
-    # Stem changes.
-    ('child', 'children'),
-    ('goose', 'geese'),
-    ('man', 'men'),
-    ('woman', 'women'),
-    ('tooth', 'teeth'),
-    ('foot', 'feet'),
-    ('mouse', 'mice'),
-    ('person', 'people'),
-    ('penny', 'pence'),
-    # Irregular suffixes.
-    ('child', 'children'),
-    ('ox', 'oxen'),
-    # f -> v.
-    ('wife', 'wives'),
-    ('wolf', 'wolves'),
-    # A few Greek and Latin plurals.
-    ('analysis', 'analyses'),
-    ('ellipsis', 'ellipses'),
-    # More exceptions
-    ('bus', 'busses'),
-    ('fez', 'fezzes'),
-    ('photo', 'photos'),
-    ('piano', 'pianos'),
-    ('halo', 'halos'),
-])
+# Load exceptions from the file.
+_exceptions = p.string_file(os.path.normpath('../../data/en/exceptions.tsv'))
 
 # If a singular noun ends in -y and the letter before the -y is a consonant, change the ending to -ies.
 # city - cities
 # puppy - puppies
 _ies = _sigma + _c + p.cross('y', 'ies')
 
-# If the singular noun ends in -on, the plural ending is usually -a.
-# phenomenon - phenomena
-# criterion - criteria
-_a = _sigma + p.cross('on', 'a')
-
-# If the singular noun ends in -us, the plural ending is frequently -i.
-# cactus - cacti
-# focus - foci
-_i = _sigma + p.cross('us', 'i')
-
-# These three can act in parallel, no conflicts.
-_ies_a_i = p.union(_ies, _a, _i)
-
-# If the singular noun ends in -s, -ss, -sh, -ch, -x, -o or -z, you usually add -es to the end.
-# iris - irises
+# If the singular noun ends in -s, -ss, -sh, -ch, -x or -z, you usually add -es to the end.
 # truss - trusses
 # marsh - marshes
 # lunch - lunches
 # tax - taxes
 # blitz - blitzes
-# potato - potatoes
 
 # Exceptions:
+# iris - irises
 # bus - busses
 # fez - fezzes
-_es = _sigma + p.union('s', 'ss', 'sh', 'ch', 'o', 'x', 'z') + pynutil.insert('es')
+_es = _sigma + p.union('ss', 'sh', 'ch', 'x', 'z') + pynutil.insert('es')
 
 # To make regular nouns plural, add -s to the end.
 # cat - cats
@@ -93,7 +49,7 @@ _s = _sigma + pynutil.insert("s")
 # Order is important, as _s rule should come in the end (catch all).
 # _exceptions->(_ies, _a, _i)->_es->s
 _plural = _priority_union(
-   _exceptions, _priority_union(_ies_a_i, _priority_union(_es, _s, _sigma),
+   _exceptions, _priority_union(_ies, _priority_union(_es, _s, _sigma),
                                 _sigma), _sigma).optimize()
 
 def inflect(singular: str) -> str:
--- a/fst/en/inflection.py
+++ b/fst/en/inflection.py
@@ -1,0 +1,100 @@
+"""
+Inflects English nouns - creates a plural form from singular.
+
+Install Pynini package before running.
+"""
+
+import pynini as p
+
+from pynini.lib import pynutil
+from pynini.lib import rewrite
+
+def _priority_union(q: p.Fst, r: p.Fst, sigma: p.Fst) -> p.Fst:
+  complement_domain_q = sigma - p.project(q, 'input')
+  return p.union(q, complement_domain_q @ r)
+
+_v = p.union('a', 'e', 'i', 'o', 'u')
+_c = p.union('b', 'c', 'd', 'f', 'g', 'h', 'j', 'k', 'l', 'm', 'n',
+             'p', 'q', 'r', 's', 't', 'v', 'w', 'x', 'y', 'z')
+_sigma = p.union(_v, _c).closure().optimize()
+
+_exceptions = p.string_map([
+    # Zero plurals.
+    'deer',
+    'fish',
+    'sheep',
+    'series',
+    'species',
+    # Stem changes.
+    ('child', 'children'),
+    ('goose', 'geese'),
+    ('man', 'men'),
+    ('woman', 'women'),
+    ('tooth', 'teeth'),
+    ('foot', 'feet'),
+    ('mouse', 'mice'),
+    ('person', 'people'),
+    ('penny', 'pence'),
+    # Irregular suffixes.
+    ('child', 'children'),
+    ('ox', 'oxen'),
+    # f -> v.
+    ('wife', 'wives'),
+    ('wolf', 'wolves'),
+    # A few Greek and Latin plurals.
+    ('analysis', 'analyses'),
+    ('ellipsis', 'ellipses'),
+    # More exceptions
+    ('bus', 'busses'),
+    ('fez', 'fezzes'),
+    ('photo', 'photos'),
+    ('piano', 'pianos'),
+    ('halo', 'halos'),
+])
+
+# If a singular noun ends in -y and the letter before the -y is a consonant, change the ending to -ies.
+# city - cities
+# puppy - puppies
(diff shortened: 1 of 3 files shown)


macchiati:
Looking good. Most common cases are covered, and can be extended.

Examples x-in-law —> xs-in-law.

This presumes that the noun is not genitive: dog's —> dog's

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+
+# If a singular noun ends in -y and the letter before the -y is a consonant, change the ending to -ies.
+# city - cities
+# puppy - puppies
+_ies = _sigma + _c + p.cross('y', 'ies')
+
+# If the singular noun ends in -on, the plural ending is usually -a.
+# phenomenon - phenomena
+# criterion - criteria
+_a = _sigma + p.cross('on', 'a')
```
grhoten, line 63: This one is sketchy. Many English words end with "tion" or "ion" that don't follow this rule.  I count about 9 in English that follow this rule, and there's greater than 2,572 that use the default rule instead.

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+  complement_domain_q = sigma - p.project(q, 'input')
+  return p.union(q, complement_domain_q @ r)
+
+_v = p.union('a', 'e', 'i', 'o', 'u')
+_c = p.union('b', 'c', 'd', 'f', 'g', 'h', 'j', 'k', 'l', 'm', 'n',
+             'p', 'q', 'r', 's', 't', 'v', 'w', 'x', 'y', 'z')
+_sigma = p.union(_v, _c).closure().optimize()
+
+_exceptions = p.string_map([
+    # Zero plurals.
```
grhoten, line 22: I count about 444 words in English that follow this rule so far. This is a good start.

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+    ('foot', 'feet'),
+    ('mouse', 'mice'),
+    ('person', 'people'),
+    ('penny', 'pence'),
+    # Irregular suffixes.
+    ('child', 'children'),
+    ('ox', 'oxen'),
+    # f -> v.
+    ('wife', 'wives'),
+    ('wolf', 'wolves'),
```
grhoten, line 43: I count about 17 words that follow this logic. Most of them are variations of compound words that end with "knife", "wife", and "life".

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+
+# If the singular noun ends in -on, the plural ending is usually -a.
+# phenomenon - phenomena
+# criterion - criteria
+_a = _sigma + p.cross('on', 'a')
+
+# If the singular noun ends in -us, the plural ending is frequently -i.
+# cactus - cacti
+# focus - foci
+_i = _sigma + p.cross('us', 'i')
```
grhoten, line 68: This is an uncommon transformation. I count about 77 nouns that follow this pattern, and about 340 nouns that do not follow this pattern.

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+# marsh - marshes
+# lunch - lunches
+# tax - taxes
+# blitz - blitzes
+# potato - potatoes
+
+# Exceptions:
+# bus - busses
+# fez - fezzes
+_es = _sigma + p.union('s', 'ss', 'sh', 'ch', 'o', 'x', 'z') + pynutil.insert('es')
```
grhoten, line 85: In my logic, I use `'ss', 'sh', 'ch', 'x', 'z'`. I count about 23 words that use the 'o' rule, and about 1,282 words that end with an 'o' that don't follow that rule for 'o'.  About 714 use the next rule for `_s`. I recommend a smaller default set.

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+    ('man', 'men'),
+    ('woman', 'women'),
+    ('tooth', 'teeth'),
+    ('foot', 'feet'),
+    ('mouse', 'mice'),
+    ('person', 'people'),
+    ('penny', 'pence'),
+    # Irregular suffixes.
+    ('child', 'children'),
+    ('ox', 'oxen'),
```
grhoten, line 40: This is a good start. The list can get much larger. You may want to make this list data driven instead of embedding it in code.

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+    ('wolf', 'wolves'),
+    # A few Greek and Latin plurals.
+    ('analysis', 'analyses'),
+    ('ellipsis', 'ellipses'),
+    # More exceptions
+    ('bus', 'busses'),
+    ('fez', 'fezzes'),
+    ('photo', 'photos'),
+    ('piano', 'pianos'),
+    ('halo', 'halos'),
```
grhoten, line 52: This list of exceptions for the "o" ending is much larger than the regular rule below. I recommend inverting this part of the logic.

fst/en/inflection.py
```diff
@@ -0,0 +1,100 @@
...
+    ('penny', 'pence'),
+    # Irregular suffixes.
+    ('child', 'children'),
+    ('ox', 'oxen'),
+    # f -> v.
+    ('wife', 'wives'),
+    ('wolf', 'wolves'),
+    # A few Greek and Latin plurals.
+    ('analysis', 'analyses'),
+    ('ellipsis', 'ellipses'),
```
grhoten, line 46: I count about 187 words that follow this pattern.

grhoten (commented):
Trick question, what's the plural of "brush"? It's both "brush" **and** "brushes".  It depends on what you're talking about.  A paint brush can become brushes, but brush in the forest is uncountable and thus it stays as brush.

This logic seems interesting. Before using it, I really want to see better vowel handling. That will typically needs Unicode normalization and a UnicodeSet. The UnicodeSet needs to handle different sets at the beginning and the end of a word for English.  Typically the letter "y" is a vowel at the end and a consonant at the beginning.

I sometimes wonder how ICU transliteration would work. I'm unsure about the performance, but it's more data driven.

fst/en/inflection.py
nciric: All of the exceptions above and the rules would be dumped into an FST file that could be used from C++ (OpenFST). Exceptions wouldn't be in the code - we use python to build the rules and serialize the "model" to disk.
Of course, we can move this list to a file and load it instead of specifying in Python code. A refactor for later.

fst/en/inflection.py
nciric: I'll move it to the exceptions list.

fst/en/inflection.py
nciric: It seems that we have a problem with Greek and Latin words and rules around them. The easiest way to just move those into exceptions list (which as you mentioned before would grow).

fst/en/inflection.py
nciric: Refactored exceptions to the file in data/ folder, but they'll still be part of the final FST once dumped.

fst/en/inflection.py
grhoten: On a related note, the plural of octopus is octopuses in English. Though it's not uncommon to see people use octopi, which is a hypercorrection.

Also see https://twitter.com/MontereyAq/status/1535752435863588866 for reference.

fst/en/inflection.py
```diff
@@ -0,0 +1,60 @@
...
+
+def _priority_union(q: p.Fst, r: p.Fst, sigma: p.Fst) -> p.Fst:
+  complement_domain_q = sigma - p.project(q, 'input')
+  return p.union(q, complement_domain_q @ r)
+
+_v = p.union('a', 'e', 'i', 'o', 'u', 'A', 'E', 'I', 'O', 'U')
+_c = p.union('b', 'c', 'd', 'f', 'g', 'h', 'j', 'k', 'l', 'm', 'n',
+             'p', 'q', 'r', 's', 't', 'v', 'w', 'x', 'y', 'z',
+             'B', 'C', 'D', 'F', 'G', 'H', 'J', 'K', 'L', 'M', 'N',
+             'P', 'Q', 'R', 'S', 'T', 'V', 'W', 'X', 'Y', 'Z')
```
grhoten, line 23: Can we replace this by a UnicodeSet and Unicode normalization to handle all of the vowels versus consonants?

Personally I split the vowel set in English to be with a y when it's a suffix, and as a consonant as a prefix.

grhoten:
> Looking good. Most common cases are covered, and can be extended. Examples x-in-law —> xs-in-law.

@macchiati I think your example is where the "x" can be replaced by a family relationship, like "brothers-in-law", right?

I've also seen "gluteus maximus" turning into "glutei maximi" for the plural form.  That's Latin in origin, and it doesn't follow English rules. It's still used by doctors and gyms.

There can also be some discussion around whether the plural of "attorney general" is "attorneys general" or "attorney generals". I suspect that such a pluralization will be an American English versus British English discussion.

grhoten (approved):
I think this is a good start. It can probably be refined further as needed, if this is the direction that we're going.

nciric:
> I think this is a good start. It can probably be refined further as needed, if this is the direction that we're going.

This is *a* proposal that has some promise. I would like to see other approaches before we proceed.

Pynini library can do many more things than I exposed in this example, also I still need to see the size of resulting grammar transition file (once you build the rules) - it could be prohibitive in size for some languages.

richgillam (commented):
Sorry I'm late to the party once again, although I don't think I have lot to add to George's comments.

This looks cool.  I'm not familiar with Panini, but the gist is fairly clear and the approach makes sense.  (My couple of nits aren't really important, of course, as this is a proof of concept.)

I'm guessing the idea here is that you'd use this tool to produce an FST that could be embedded into C code for performance, and I think that makes sense.  I do have a couple of thoughts, though.  I apologize if they're obvious to everyone else...

1. Ideally, what you'd want to do, I think, is start with a large collection of words with all of their inflected forms and glean the transformation rules by examination.  That is, rather than hand-crafting transformation rules as you've done here, you'd have some kind of tool that boils them down from looking at the patterns that actually occur in the lexicon and their frequency.  Maybe you'd have to hand-craft _some_ rules that the tool couldn't get by examination, but it seems like most of this could be derived by examination.
2. If we think about it that way, what all of this basically amounts to is a compression algorithm-- conceptually, you're just looking up inflected forms in a lexicon, and using an FST is just a way of making that lexicon smaller and faster.

As I was writing item 1, I realized I'm sort of describing a deep-learning model, but I think I'm hoping for something simpler and more debuggable that that would get us, something more akin to actually compressing the lexicon data (if that makes any sense).

The other thing we'd have to think a lot about, but maybe later, is what to do when results are ambiguous-- is there a way to tell when the plural of "person" should be "people" and when it should be "persons," for example?  Or do we return both, maybe tagged with probabilities?

As I said, this is probably all obvious to everybody reading this, and if so, I apologize...

macchiati:
https://www.merriam-webster.com/grammar/the-many-plurals-of-octopus-octopi-octopuses-octopodes

nciric:
> Sorry I'm late to the party once again, although I don't think I have lot to add to George's comments.
> 
> This looks cool. I'm not familiar with Panini, but the gist is fairly clear and the approach makes sense. (My couple of nits aren't really important, of course, as this is a proof of concept.)
> 
> I'm guessing the idea here is that you'd use this tool to produce an FST that could be embedded into C code for performance, and I think that makes sense. I do have a couple of thoughts, though. I apologize if they're obvious to everyone else...
> 
> 1. Ideally, what you'd want to do, I think, is start with a large collection of words with all of their inflected forms and glean the transformation rules by examination.  That is, rather than hand-crafting transformation rules as you've done here, you'd have some kind of tool that boils them down from looking at the patterns that actually occur in the lexicon and their frequency.  Maybe you'd have to hand-craft _some_ rules that the tool couldn't get by examination, but it seems like most of this could be derived by examination.
> 2. If we think about it that way, what all of this basically amounts to is a compression algorithm-- conceptually, you're just looking up inflected forms in a lexicon, and using an FST is just a way of making that lexicon smaller and faster.
> 
> As I was writing item 1, I realized I'm sort of describing a deep-learning model, but I think I'm hoping for something simpler and more debuggable that that would get us, something more akin to actually compressing the lexicon data (if that makes any sense).
> 
> The other thing we'd have to think a lot about, but maybe later, is what to do when results are ambiguous-- is there a way to tell when the plural of "person" should be "people" and when it should be "persons," for example? Or do we return both, maybe tagged with probabilities?
> 
> As I said, this is probably all obvious to everybody reading this, and if so, I apologize...

There are unanswered questions with my approach - multiple results for example. Also I am not 100% sure we'll be able to avoid Python logic, so in addition to FST table you'll need to replicate some of the code too (In this example we didn't have that problem).

I think rule generation can be automated to some degree using data patterns and light ML, for example clustering for y->ies, Greek/latin words etc.

As I mentioned before, I would like to learn of other approaches before we commit to any single one.

Another part I didn't do was generate actual transition table and use it from OpenFST C library and see how that works (both in size and ease of use).

PR #31, opened 2024-05-14
Adding Serbian language

nciric:


nciric:
Serbian group 1 still has some exceptions to be ironed out. I'll file an issue, but it's ready to try it out for CLDR data.

Issue #19, opened 2024-03-14
Support inflection of nouns


grhoten:
We should be able to inflect common nouns and proper nouns. This would typically include being able to modify the grammatical gender, grammatical number and grammatical cases in a lot of languages.

Prepositions in English take on grammatical case in many other languages. Typically in the form of suffixes to nouns. So this makes it related to issue https://github.com/unicode-org/inflection/issues/17.

English possessive/genitive forms of nouns typically need to add `'s` or just `'`, but that algorithmic logic is a lot harder in other languages, like German, Danish, Dutch, Russian and so forth.

For example, you should be able to turn "city" into "city's" or turn "cities" into "cities'". For a language like Russian, you can look at [кот](https://www.wikidata.org/wiki/Lexeme:L31687) for an example.

Here's a more compact declension table for looking at such information.
cat (кот) | singular | plural
-- | -- | --
nominative | кот | коты
genitive | кота | котов
accusative | коту | котам
dative | кота | котов
instrumental | котом | котами
prepositional | коте | котах

nciric:
+1

I think this was a main use case when we started discussing this project as most of the placeholders in messages are nouns (proper or common).

The solution will probably range from simple/complex algorithms + lexicon exceptions, to potentially ML models for some languages. I feel this is the first problem we should tackle, as it intersects well with common needs.

grhoten:
I say that for single words or very few words, ML is undesirable. From experience, it’s very resource intensive, which makes it undesirable for resource constrained environments. There are many languages where a traditional algorithmic solution for out of vocabulary words is cheaper, faster, smaller, quicker to implement and more accurate than an ML solution. I have some horror stories around this topic.

If you start handling many words or a whole sentence, ML starts looking more appealing because such solutions thrive on context.

I’d say the only exception to this rule are agglutinative languages, like Finnish and perhaps Turkish. A generally ML approach is more likely accurate in such languages. That requires a lengthier overview and education session on the topic.

The ML versus rule based approach will probably involve a discussion to find the right balance.

nciric:
> I’d say the only exception to this rule are agglutinative languages, like Finnish and perhaps Turkish. A generally ML approach is more likely accurate in such languages. That requires a lengthier overview and education session on the topic.

I expect most languages will be fine with the algorithmic + lexicon approach. I would use ML only when necessary, as you mentioned in Finnish/Turkish. So this is not a decision we  need to make a head of time, just a reminder that we need to organize our code to allow different implementations.

[...]

Issue #18, opened 2024-03-14
Support metadata with vowel and consonant or pronunciation properties


grhoten:
There are several languages that change their choice of articles or prepositions depending on whether it starts or ends in a vowel or consonant.  As an example, the word "[apple](https://www.wikidata.org/wiki/Lexeme:L3257)" has IPA information for pronouncing words in Wikidata. The vowel and consonant properties can be derived from that information. Properly supporting the English indefinite article requires this information to handle all of the edge cases. For example, in English you say "an apple" and not "a apple". You can make default guesses with a UnicodeSet to check the base character being in "[aeiou]" for the front of the word, but you have to handle such edge cases with exceptions, such as "an LED light", or "a unicorn".

grhoten:
I'd like to nominate this to be resolved with pull request #35. Though some additional work may be needed. The functionality is there, but the Wikidata lexeme ingestion will likely need to derive this data for edge cases too.

Issue #17, opened 2024-03-14
Support inflection of adpositions, prepositions and particles


grhoten:
We should be able to inflect prepositions. These words can include "on", "in", "of" and other English equivalents. French, Spanish, Italian, Portuguese, Russian and other languages have prepositions that vary depending on the noun being referenced.

For example, in French, I can use "d’" or "de" depending on whether the next word starts with a vowel or a consonant.

For a language like Korean, it's whether the the previous word ends with a vowel, consonant or rieul. For example, "house" ends with a consonant instead of a vowel.

For some languages, prepositions can combine with definite articles. For example, Spanish has "del", "de la", "de los" and "de las". French has another example with "au ", "à la ", "à l’" and "aux ".

grhoten:
I'd like to nominate this to be resolved with pull request #35.

Issue #15, opened 2024-03-14
Add support for measure words or counter words or classifiers


grhoten:
This is mostly an East Asian language topic. I'm listing this topic as something to consider. I'm OK with not supporting this functionality, since it can be supported separately from a structured lexicon.

For the longer description, please see https://en.wikipedia.org/wiki/Classifier_(linguistics) for more details.

grhoten:
I'd like to nominate this to be resolved with pull request #35.

PR #35, opened 2024-10-30
Contribute Morphuntion from Apple as open source software

grhoten:
This contribution is related to the [Unicode Technical Workshop topic about Inflection](https://www.youtube.com/watch?v=Z9M8SwIbvMM).

This contribution should resolve the following issues: #5, #6, #7, #11, #12, #13, #15, #17, #18, #19
This contribution is also related to the following issues without fully resolving the issues: 3, 4, 8, 10, 21, 23, 24, 25
This contribution also has an implementation that addresses these CLDR issues: 13025, 13563


morphuntion/resources/org/unicode/morphuntion/dictionary/dictionary_ar.lst
```diff
@@ -0,0 +1,63 @@
+مطعم: singular genitive nominative indefinite construct masculine participle active passive adjective noun verb inflection=7
```
nciric, line 1: Similar to contractions comment - we should have a better explanation of fields. Do they need to be in some specific order? What are the possible values? Point to tools used to make them?

morphuntion/resources/org/unicode/morphuntion/inflection/pronoun.properties
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright 2022-2024 Apple Inc. All rights reserved.
+#
+yue_HK=yue_Hant
```
nciric, line 4: Is this just mapping one form of LangID to another? Why (comment)?

morphuntion/resources/org/unicode/morphuntion/tokenizer/config_ar.properties
```diff
@@ -0,0 +1,13 @@
...
+tokenizer.implementation.class=locale.ar.ArTokenizer
+tokenizer.nonDecompound.file=/org/unicode/morphuntion/tokenizer/ar/nondecompound.tok
+tokenizer.prefixes.0=\u0648,\u0641
+tokenizer.prefixes.1=\u0628,\u0643,\u0644
+tokenizer.prefixes.2=\u0627\u0644,\u0644
+tokenizer.suffixes=\u064A,\u0646\u064A,\u0643\u064E,\u0643\u0650,\u0643,\u0647,\u0647\u0627,\u0646\u0627,\u0643\u0645\u0627,\u0647\u0645\u0627,\u0646\u0627,\u0643\u0645,\u0643\u0646,\u0643\u0648,\u0647\u0645,\u0647\u0646,\u0647\u0645
+tokenizer.normalize=\u062A
+tokenizer.replace=\u0629
+
+#tokenizer.decompound=(\u0648|\u0641)?(\u0628|\u0643|\u0644)?(\u0627\u0644)?(.+?)(\u064A|\u0646\u064A|\u0643|\u0646\u0627|\u0643\u0645|\u0643\u0646|\u0647|\u0647\u0627|\u0643\u0646|\u0647\u0645|\u0647\u0646)?
```
nciric, line 13: Remove?

morphuntion/src/icu4cxx/BreakIterator.hpp
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2016-2024 Apple Inc. All rights reserved.
+ */
+// This is a generated file. Do not edit this file.
```
nciric, line 4: Do we care how it's generated? If yes, some epointers to the generator would be good.

morphuntion/src/icu4cxx/BreakIterator.hpp
```diff
@@ -0,0 +1,126 @@
...
+#include <morphuntion/exception/NullPointerException.hpp>
+#include <unicode/ubrk.h>
+
+namespace icu4cxx {
+class BreakIterator {
+public:
+BreakIterator(UBreakIteratorType type, const char * locale, std::u16string_view text) {
+    UErrorCode ec = U_ZERO_ERROR;
+    wrappee_ = ubrk_open(type, locale, (const UChar *)text.data(), (int32_t)text.size(), &ec);
+    ::morphuntion::exception::ICUException::throwOnFailure(ec);
```
nciric, line 17: Some companies don't allow exceptions in code. Do these bubble up to top or are they always caught and then API returns an error?

morphuntion/src/morphuntion/Object.hpp
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2016-2024 Apple Inc. All rights reserved.
+ */
+#pragma once
```
nciric, line 4: Hopefully this works everywhere we care about (we use guards instead). https://en.wikipedia.org/wiki/Pragma_once

nciric (commented):
First round. I'll go into actual code in the next.

morphuntion/README.md
```diff
@@ -0,0 +1,96 @@
+<!--
+ Copyright 2016-2024 Apple Inc. All rights reserved.
+-->
+# Morphuntion
```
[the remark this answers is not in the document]
grhoten: We anticipated that this will likely be renamed.

morphuntion/build.gradle
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright 2020-2024 Apple Inc. All rights reserved.
+ */
+
+plugins {
+    id 'net.ltgt.errorprone' version '1.1.1' apply false
+    id 'maven-publish'
```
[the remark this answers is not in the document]
grhoten: It's dictionary-parser.  Denny will likely want to look at that code. It can be reimplemented in another programming language when using wikidata as a data source, but I wanted to provide the tool source code so that the logic behind generating the lexical dictionaries is a lot clearer.  The tool will likely be deleted once wikidata can be ingested.

This also had a Java wrapper through JNI. That implementation was excluded from this contribution at this time to expedite making this implementation open source.

morphuntion/resources/org/unicode/morphuntion/contraction/contractionExpandingTable_es.csv
```diff
@@ -0,0 +1,2 @@
+del,de,el
```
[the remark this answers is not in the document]
grhoten: The first column is the contraction. The variable number of words after the first column are what the words expand to.  At least one language depends on this to when prepositions and articles combine in variable ways.

In this scenario, "del" is the contraction of "de" and "el" (of and the).

A readme can be added, but it should be at the top level.  Any files added in these subdirectories tend to be automatically compiled or distributed.  How much detail are you looking for?  Is a couple of sentences for each data type sufficient for now?

morphuntion/resources/org/unicode/morphuntion/dictionary/dictionary_ar.lst
grhoten: The dictionary-parser historically generates these files, but in this contribution, it's only helpful as a reference. It will likely need to be replaced by a tool that generates equivalent data from Wikidata.

Everything before the colon is the word/phrase.  Everything after the colon and without an = is a white space separated list of grammemes. Anything after the colon is a list of key/value pairs. The key is not unique.  Think of it as a multi-value map.  Typically it's the key "inflection" with a value of the inflection table name.  The dictionary-parser typically enumerates the name. You can hand craft the name, but it's a pain to maintain such names.

morphuntion/resources/org/unicode/morphuntion/inflection/pronoun.properties
grhoten: It's an aliasing scheme. In this case, we're making the script explicit for a given region.

morphuntion/src/icu4cxx/BreakIterator.hpp
grhoten: This is generated with the generateICUWrapper tool.  It takes an existing ICU installation, and generates C++ wrappers around the C API.  This provides a natural adaptor for C++ API, but with the stability of the C ABI.

morphuntion/src/icu4cxx/BreakIterator.hpp
grhoten: They're not caught on the C++ API. The C API wrappers around the C++ code converts the exceptions to CFErrorRef.  If you don't want C++ exceptions, use the C API.  If you want C++ exceptions, use the C++ API.  C++ library code also generates exceptions. So you can't completely omit them from C++.

morphuntion/src/morphuntion/Object.hpp
grhoten: That page doesn't show any compilers that don't support it.  I guess the only issue would be if you want to use an old compiler.  From a maintenance point of view, this is much more reliable than using macros.

morphuntion/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+include(GNUInstallDirs)
+
+# optionally compile with code profiling
+if(PROFILING)
+    message("-- PROFILING TURNED ON")
+    add_compile_options(-g -fprofile-instr-generate -fcoverage-mapping)
+    add_link_options(-g -fprofile-instr-generate -fcoverage-mapping)
+endif()
+
+#Set these warning properties on a project level
```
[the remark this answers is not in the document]
grhoten: I have changed this file to make the comments more consistent in style.

morphuntion/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+- Word inflection of a word into another [surface form](https://en.wikipedia.org/wiki/Surface_form) of another word.
+- Grammatical agreement between words.
+- Querying grammatical properties as [grammemes](https://en.wiktionary.org/wiki/grammeme) (the values of [grammatical categories](https://en.wikipedia.org/wiki/Grammatical_category)) 
+
+It uses C++20, [ICU4C](https://icu.unicode.org/), UTF-16 strings (just like Java) and a data source of lexical
+dictionaries that contain relationships between inflections of a word. Just like ICU, it is thread safe between service
+objects, but mutable objects are not necessarily thread safe between threads.
+
+By making this implementation open sourced, various software frameworks can generate grammatically correct messages
```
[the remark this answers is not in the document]
grhoten: I have finished this thought. I'm sure that the new sentence can be refined further later.

morphuntion/resources/org/unicode/morphuntion/dictionary/dictionary_ar.lst
grhoten: Related to this topic, any tool that generates this file will likely need a good way to filter the data.  Certain surface forms may be possible to automatically generate the data, like the English possessive/genitive form. **Most** abbreviations are unhelpful.  So you can omit them from the data.  Phrases are generally unhelpful, like "to the moon" in English.  Words that are technically correct, but are not widely understood outside the scope of a joke, like the singular form of sheep being shoop.

The order of the words don't matter, but it's easier to search the list when it's sorted.  When there are duplicate words, the second form overrides the first set of grammemes, except the key/value pairs.  For the key/value pairs, it prefixes them to the start of the set.  Typically the most popular key/value pair should be first.  So if there are 2 ways to interpret a word, like the noun "brush", the first one wins. When there is the same word, but the inflection tables have different parts of speech, usually the noun wins, but that can be overridden.  For the words "work" and "works", this distinction is important. See [morphuntion/test/resources/morphuntion/dialog/inflection/en.xml](https://github.com/unicode-org/inflection/pull/35/files#diff-47d33bf0724ecc058769400b4c5f702e9e058b9cbf7259db8b30d45871b4cbf7) lines 156-166 for an example of the caller overriding the inflection behavior for the given context.

nciric (approved):
GitHub makes reviewing 800+ files PR a nightmare (everything is super laggy and slow). Let's land this and then go over important details in follow up PRs.
I guess current legal discussion is preventing the merge.

morphuntion/tools/generateICUWrapper/generateICUWrapper.cpp
```diff
@@ -0,0 +1,551 @@
+//
+//  main.cpp
+//  icu4cxx_generate
```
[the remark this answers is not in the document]
grhoten: Yes, rewriting this in another programming language is fine. It’s rarely run anyway.

morphuntion/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
+#
+# Copyright 2018-2024 Apple Inc. All rights reserved.
```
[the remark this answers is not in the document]
grhoten: As per the email discussion, the previously included license as been removed in order to use the existing license in this repository, but these copyright notices should not be disturbed.

Issue #38, opened 2024-12-16
Morphuntion won't compile/link on Ubuntu 22


nciric:
Command:

```
make -j 8 check
```

Error message:

```
In file included from /home/cira/code/inflection/morphuntion/src/morphuntion/dictionary/PhraseProperties.cpp:11:
In file included from /home/cira/code/inflection/morphuntion/src/morphuntion/util/UnicodeSetUtils.hpp:8:
/home/cira/code/inflection/morphuntion/src/icu4cxx/UnicodeSet.hpp:163:12: error: use of undeclared identifier 'uset_hasStrings'
    return uset_hasStrings(wrappee_);
           ^
/home/cira/code/inflection/morphuntion/src/icu4cxx/UnicodeSet.hpp:184:12: error: use of undeclared identifier 'uset_getRangeCount'
    return uset_getRangeCount(wrappee_);
           ^
2 errors generated.
make[3]: *** [src/CMakeFiles/morphuntion_tool_objs.dir/build.make:188: src/CMakeFiles/morphuntion_tool_objs.dir/morphuntion/dictionary/PhraseProperties.cpp.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [CMakeFiles/Makefile2:773: src/CMakeFiles/morphuntion_tool_objs.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:944: test/CMakeFiles/check.dir/rule] Error 2
make: *** [Makefile:456: check] Error 2
```

Steps:
* ICU was compiled from sources and installed into /usr/local/lib/icu
* ICU version was either  76.1 (official), or 77.0.1 (live)
* CMake 3.25.1 or 3.31 installed
* Clang 14.0.6 with details below

```
Debian clang version 14.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/10
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/12
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/12
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

This same error shows up on ChromeBook Linux and clean Hyper-V VM of Ubuntu 22.
It's interesting that other uses of the same header in the code  base don't show up as errors.

nciric:
Cmake --version output:

```
[...]

PR #39, opened 2024-12-18
Expanding build documentation

nciric:
Expanding README.md  with pointers to ICU4C, ICU_ROOT path rules etc. and restructuring the build section.


nciric:
> Does this resolve issue #38?

No I still have the same issue. But it clarifies how to get ICU4C, and tips we discovered in #38 .

PR #40, opened 2024-12-18
Adding support for sr-SR

nciric:
Starting a how to add a new language document. I'll add more steps as I progress (I would like to  have working build before I go any deeper).

I added sr, sr_RS to the LangUtil to try it out. I see that Croatian is already there, but also not supported so I feel this is fine.


grhoten:
Please consider the following patch as a starting point.
[add-sr.patch](https://github.com/user-attachments/files/18194816/add-sr.patch)

Here are some caveats.

1. It's missing the lexical dictionary.  That's your biggest blocker.
2. The script of the pronouns (Latin) does not match the script of the RBNF rules (Cyrillic). Feel free to change as appropriate, or switch to hr to match the CLDR default.  If not that, then consider switching to sr_Latn_RS instead.
3. The RBNF rules look incomplete in CLDR.  So grammatical correctness of quantities is probably incorrect. See RuCommonConceptFactory on a more complete solution.
4. The display function needs to be implemented. See the Russian display function for a more complete implementation.
5. If there was a lexical dictionary, then it would be possible to detect the properties for grammatical number and gender as this patch is currently written.
6. I'm assuming that the properties in grammar.xml is complete enough. Feel free to verify.

nciric:
> Please consider the following patch as a starting point. [add-sr.patch](https://github.com/user-attachments/files/18194816/add-sr.patch)

Great patch, thanks!

> Here are some caveats.
> 
> 1. It's missing the lexical dictionary.  That's your biggest blocker.

 I didn't notice Croatian was already in. Do we have a lexicon for it already, or at least a test sized one (I can transliterate it)?

> 2. The script of the pronouns (Latin) does not match the script of the RBNF rules (Cyrillic). Feel free to change as appropriate, or switch to hr to match the CLDR default.  If not that, then consider switching to sr_Latn_RS instead.

Changed to Cyrillic.

> 3. The RBNF rules look incomplete in CLDR.  So grammatical correctness of quantities is probably incorrect. See RuCommonConceptFactory on a more complete solution. You may want to consider adding some test cases with different grammatical cases in QuantifyTest#testSerbian. You may find inspiration in the QuantifyTest#testRussian test for how to write such a test.

I need working build to start testing. Is it ok to submit as is for now and  come back to testing later?

> 4. The display function needs to be implemented. See the Russian display function for a more complete implementation.
> 5. If there was a lexical dictionary, then it would be possible to detect the properties for grammatical number and gender as this patch is currently written.
> 6. I'm assuming that the properties in grammar.xml is complete enough. Feel free to verify.

They seem ok, as they are based on Croatian entry already there.

> 7. It's missing some data driven tests. I recommend adding some in morphuntion/test/resources/morphuntion/dialog/inflection/sr.xml.

Is it ok to add them later after we figure out the build problem? But definitely on my TODO list.

grhoten:
After some thought, I think it's best to separate out the instructions from the patch to add a new language. The tests currently pass, and these changes are incomplete (primarily due to the missing lexical dictionary). I think the patch to add a new language needs further work, but it's a good start.

nciric:
> After some thought, I think it's best to separate out the instructions from the patch to add a new language. The tests currently pass, and these changes are incomplete (primarily due to the missing lexical dictionary). I think the patch to add a new language needs further work, but it's a good start.

I moved the instruction doc to a separate PR (you are on it).

nciric:
@grhoten I have two question that would speed things up:
*  Where do I implement inflection rules? Currently the tests fail with "Inflector not found for sr" message.
*  What's the process of adding a dictionary?
   * I can craft a manual one (to pass the tests) following the format for Russian inflection/resources/org/unicode/inflection/dictionary/dictionary_ru.lst
   * But how do I generate the rules xml file (inflection/resources/org/unicode/inflection/dictionary/inflectional_ru.xml) and then update the lexicon entries with inflection= value?

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -0,0 +1,64 @@
...
+    ::std::u16string displayString = displayValue->getDisplayString();
+    if (displayString.empty()) {
+        return nullptr;
+    }
+
+    // To make compiler quiet about unused variable.
+    if (enableInflectionGuess)
+        return nullptr;
+
+    // TODO Implement the rest
```
grhoten, line 60: Do you have plans to implement this TODO?

nciric: I still have some tests failing. My goal is to have tests pass, submit this PR, then open a new one with affecting only this file.
WDYT?

grhoten: That sounds like a good plan.

PR #41, opened 2024-12-19
Separate how to add  a new language from code change.

nciric:
Separated how to from the code changes.


nciric:
We discussed in 2025-01-07 meeting that this can be committed so that others can contribute. I will add more to the document as I work on PR #40 .

Issue #42, opened 2025-01-09
Rename Morphuntion to Unicode Inflection and reduce CoreFoundation dependency


grhoten:
As per the discussion in the last Inflection meeting (1/7/2025), the decision was to rename Morphuntion to the Unicode Inflection library.

There was also a decision to reduce the usage of the CoreFoundation. This change will remove the CFLocaleRef and CFErrorRef usage with the rest of it being handled with #36.  CFLocaleRef will be switch to const char*, and CFErrorRef will switch to UErrorCode.

This issue for this pull request will address both of these decisions.

Issue #36, opened 2024-12-10
Remove the Core Foundation dependency in Morphuntion


grhoten:
It would be helpful to reduce the size of the dependencies and to be able to compile the API with the gcc compiler. This dependency has a Clang compiler dependency because the gcc compiler doesn't define the endianness that the Core Foundation headers will accept. This dependency is only needed for the C API. Perhaps we can consider mapping these types to something that ICU4C is more closely aligned with.

When mapping the data types, we should consider the following questions:

- How should informative error message in exceptions be conveyed?  Use UErrorCode instead and lose the diagnostic information?
- How should a locale be specified? const char *?
- How should a string be specified? const char16_t * with length parameter? The code currently requires C11 as a minimum.
- How should an array or a map of objects be represented?  This is a less trivial of a mapping, but it’s doable.

The only exception at this time is the UTF-16 data type. The current code depends on C11, which needs it to be char16_t instead of UChar, which is uint16_t in ICU4C.

nciric:
I think we can close this one.

PR #43, opened 2025-01-09
Inflection-42 Rename Morphuntion to Unicode Inflection and reduce CoreFoundation dependency

grhoten:
This resolves Issue #42.

As per the discussion in the last Inflection meeting (1/7/2025), the decision was to rename Morphuntion to the Unicode Inflection library.

There was also a decision to reduce the usage of the CoreFoundation. This change will remove the CFLocaleRef and CFErrorRef usage with the rest of the changes being handled with https://github.com/unicode-org/inflection/issues/36. CFLocaleRef will be switch to const char*, and CFErrorRef will switch to UErrorCode in this pull request.

This pull request built and passed the tests on macOS and Linux. Valgrind reported no memory issues. The check-headers target reported no issues.


nciric (approved):
Since we don't have CI yet I will assume that the local build/test is passing for you.

nciric:
Could you rename paths in [how-to-add-language document](https://github.com/unicode-org/inflection/blob/main/documents/how_to_add_new_language.md) too? In this or separate PR.

grhoten:
I'll make the change in this pull request.

PR #74, opened 2025-01-28
Expand Serbian lexicon for wikidata

nciric:
Adding ~150 nouns so we can run the tools. I still need to add pronouns etc.


nciric:
This is just a local change related to Serbian tools. I'll just merge so that Denny can grab data and add to Wikidata.

PR #92, opened 2025-03-07
Adding git lfs support guidance.

nciric:


inflection/README.md
```diff
@@ -80,6 +80,19 @@ The following are the dependencies to use this code.
...
+Additional checkout steps are necessary when working with the repository as it utilizes Git LFS files.
+
+##### Get LFS client and install it, e.g. on Ubuntu
+
+```bash
+sudo apt install git-lfs
+git lfs install
+```
+
+For more details and troubleshooting refer to this [answer](https://stackoverflow.com/questions/72610494/what-is-the-difference-between-git-lfs-fetch-git-lfs-fetch-all-and-git).
```
grhoten, line 94: Can we use a more official URL? For macOS, I go to https://git-lfs.com/ instead. I've never encountered the issues that were mentioned in the thread that you referenced, which be more confusing than necessary.

nciric: I had some branches created before LFS was added and plain instructions didn't "take". I had to do "git LFS checkout" in the branch to get LFS to actually link files.
I too was not very happy with stack overflow link, and given that new contributors won't have my problems, pointing to official documentation is probably enough.

PR #111, opened 2025-04-30
Landing sr support for dictionary words

nciric:
This implements dictionary lookup from Serbian lexicon.

Some words fail to properly inflect, even though they should match inflection rules, e.g.:
1. уранак should inflect by group  f  (see word пропланак)
2. игроказ should inflect by group 29 (see word путоказ)
3. пашњак  (maybe special case as it inserts additional a to the suffix)

Those examples are currently commented out.


nciric:
George replied offline with:

If the word is not in the dictionary, then you need to build a heuristic for handling such situations.  See EnGrammarSynthesizer_EnDisplayFunction::guessPluralInflection for an example way to guess words.

Your current code only works for words in the lexical dictionary, and the code can take some ambiguity into account when doing the inflection table lookup.

Issue #118, opened 2025-05-27
Add memory leak checking to pull request checks


grhoten:
We should add leak checking to the Linux pull request GitHub workflow. The easy method for doing that is to use valgrind.

Something like the following command should be used.

```
make check TEST_RUN_COMMAND="valgrind --leak-check=full --show-leak-kinds=definite --errors-for-leak-kinds=definite --error-exitcode=1"
```

PR #121, opened 2025-06-04
Add CI for valgrind fix issues/118

FrankYFTang:
Fix https://github.com/unicode-org/inflection/issues/118

### DIFF
--- a/.github/workflows/ubuntu-memory-check.yml
+++ b/.github/workflows/ubuntu-memory-check.yml
@@ -1,6 +1,4 @@
-# This starter workflow is for a CMake project running on multiple platforms. There is a different starter workflow if you just want a single platform.
-# See: https://github.com/actions/starter-workflows/blob/main/ci/cmake-single-platform.yml
-name: CMake on multiple platforms
+name: Check memory leak by using valgrind
 
 on:
   # Keep it for manual runs.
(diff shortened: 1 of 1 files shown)


FrankYFTang:
macOS normail build CI take 2m5s to run, ubuntu take 3m24s to run, checkmemleak take 5m25s to run

FrankYFTang:
@nciric Please review

FrankYFTang:
Fix #118

FrankYFTang:
See https://github.com/FrankYFTang/inflection/actions/runs/15430643627/job/43427674705 for the test run

nciric (commented):
I see two ways to go here:
1. Remove duplication of ICU checkout and some checkout steps (we are potentially paying 2x for LFS checkout) and keep everything in this file
2. Create a new workflow (e.g. ubuntu-memory-check.yml) and move your valgrind logic there. You will need to copy some of the triggering logic from this file.

FrankYFTang:
> I see two ways to go here:
> 
> 1. Remove duplication of ICU checkout and some checkout steps (we are potentially paying 2x for LFS checkout) and keep everything in this file
> 2. Create a new workflow (e.g. ubuntu-memory-check.yml) and move your valgrind logic there. You will need to copy some of the triggering logic from this file.

I address both issues . PTAL

PR #123, opened 2025-06-05
Added pronouns for Malayalam

deonajulary06:
Added pronouns for Malayalam to help with grammar and inflection


deonajulary06:
@grhoten One of the checks I'm failing looks like it's because the tester's ICU version is too low (ICU version found 74.2. Version 77.1 is the minimum acceptable version). Is this something I fix on my end or Frank's end?

grhoten:
I think the version check problem is something for @FrankYFTang to address. I don’t think that it’s an issue in this pull request.

PR #128, opened 2025-06-10
Cache LFS content (fix issue/127)

### LINKED ISSUE #127: cache lfs in github action
@nciric  mentioned that we currently checkout with lfs everytime and it is better to cache the lfs content instead of automatically sync it. 

The way to address that is stated in https://github.com/actions/checkout/issues/165#issuecomment-2776048200

We should try it with our build to reduce bandwidth on pulling lfs

FrankYFTang:
Fix #127

### DIFF
--- a/.github/workflows/cmake-multi-platform.yml
+++ b/.github/workflows/cmake-multi-platform.yml
@@ -90,6 +90,9 @@ jobs:
           ${{ runner.os }}-lfsdata-v1-
           ${{ runner.os }}-lfsdata
 
+    - name: Check LFS restore the files or not
+      run: ls -l inflection/resources/org/unicode/inflection/dictionary/*
+
     - name: Fetch any needed Git LFS objects and prune extraneous ones
       if: steps.lfscache.outputs.cache-hit != 'true'
       run: git lfs fetch --prune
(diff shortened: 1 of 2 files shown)


FrankYFTang:
Test run on 
https://github.com/FrankYFTang/inflection/actions/runs/15597515786
and
https://github.com/FrankYFTang/inflection/actions/runs/15597601844

.github/workflows/cmake-multi-platform.yml
```diff
@@ -71,8 +76,24 @@ jobs:
...
+        path: .git/lfs
+        key: ${{ runner.os }}-lfsdata-v1-${{ hashFiles('.git/lfs-hashes.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-lfsdata-v1-
+          ${{ runner.os }}-lfsdata
+
+    - name: Fetch any needed Git LFS objects and prune extraneous ones
+      run: git lfs fetch --prune
+
+    - name: Check out Git LFS content
```
nciric, line 95: I looked at the output (your test runs) and it seems that we had a cache hit, but this line still outputs:

```
Checking out LFS objects: 100% (8/8), 107 MB | 0 B/s, done
```

Is that expected?

FrankYFTang: ok, I need to check the if

here are the new run
https://github.com/FrankYFTang/inflection/actions/runs/15622526187

and
https://github.com/FrankYFTang/inflection/actions/runs/15622524628

nciric (approved):

PR #129, opened 2025-06-11
Create distribution packages for ubuntu (.deb release v0.0.1)

dockerh:
##  Ubuntu Packaging with CPack & GitHub Actions

This PR introduces a GitHub Actions workflow for building and packaging the Unicode Inflection library on Ubuntu using CPack. It generates `.deb` and `.tar.gz` artifacts and uploads them to GitHub Releases when a Git tag is pushed. ICU 77.1 is installed from a prebuilt binary release.

---

####   CMake / CPack Configuration
- Updates `CMakeLists.txt` to:
  - Dynamically parse version from Git tags (e.g. `Inflection-0.1.0`)
  - Configure CPack to generate `.deb` and `.tar.gz` artifacts

####  GitHub Actions Workflow
- Workflow file: `.github/workflows/create-ubuntu-distribution-packaging.yml`
- Triggered on tag push (e.g., `v0.1.0`)
- Key steps:
  - Installs ICU 77.1 from a prebuilt Ubuntu binary
  - Builds using Clang with CMake
  - Runs tests with `make check`
  - Packages using `cpack`
  - Uploads artifacts to GitHub Release

---

###  Outputs

-  Ubuntu `.deb` package
-  Source `.tar.gz` package

### DIFF
--- a/.github/workflows/create-ubuntu-distribution-packaging.yml
+++ b/.github/workflows/create-ubuntu-distribution-packaging.yml
@@ -1,3 +1,12 @@
+# ----------------------------------------------------------------------------------------
+# To update ICU version:
+# 1. Update the ICU_VERSION environment variable below.
+# 2. Check the ICU binary URL — is the filename still Ubuntu22.04-x64.tgz?
+#    (e.g., icu4c-78_1-Ubuntu22.04-x64.tgz)
+# 3. Update inflection/CMakeLists.txt:
+#    - Modify `CPACK_DEBIAN_PACKAGE_DEPENDS` to match the new ICU version.
+# ----------------------------------------------------------------------------------------
+
 name: Build & Package (Ubuntu)
 
 on:
--- a/inflection/CMakeLists.txt
+++ b/inflection/CMakeLists.txt
@@ -136,8 +136,10 @@ install(DIRECTORY ${INFLECTION_DATA_ROOT}/    TYPE DATA    COMPONENT inflection_
 # CPack Configuration for Ubuntu Packaging
 
 set(CPACK_PACKAGE_NAME "unicode-inflection")
-set(CPACK_PACKAGE_VERSION "${INFLECTION_VERSION}")
 
+# Apply the current tagged Inflection version to the CPack version.
+# CPack may inherit a default or cached version, so we explicitly set it here.
+set(CPACK_PACKAGE_VERSION "${INFLECTION_VERSION}")
 
 string(REPLACE "." ";" INFLECTION_VERSION_LIST "${INFLECTION_VERSION}")
 
@@ -156,8 +158,6 @@ endif()
 if(_len GREATER 2)
   list(GET INFLECTION_VERSION_LIST 2 CPACK_PACKAGE_VERSION_PATCH)
 endif()
-
-
 
 set(CPACK_PACKAGE_VENDOR "Unicode Consortium")
 set(CPACK_PACKAGE_CONTACT "https://github.com/unicode-org/inflection")


inflection/CMakeLists.txt
```diff
@@ -130,3 +130,23 @@ make generate-coverage-csv  : Generates code coverage as a csv\\n\
 install(TARGETS   inflection                  LIBRARY      COMPONENT inflection_library)
 install(DIRECTORY ${INFLECTION_INCLUDE_ROOT}/ TYPE INCLUDE COMPONENT inflection_headers)
 install(DIRECTORY ${INFLECTION_DATA_ROOT}/    TYPE DATA    COMPONENT inflection_data)
+
+# CPack Configuration for Ubuntu Packaging
+
+set(CPACK_PACKAGE_NAME "unicode-inflection")
+set(CPACK_PACKAGE_VENDOR "Unicode Consortium")
+set(CPACK_PACKAGE_CONTACT "grhoten@gmail.com")
```
grhoten, line 138: I think we need to find another contact. I'm not sure that I'm ready to have my email address here.

inflection/CMakeLists.txt
```diff
@@ -130,3 +130,23 @@ make generate-coverage-csv  : Generates code coverage as a csv\\n\
...
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Unicode Inflection Library")
+set(CPACK_GENERATOR "DEB;TGZ")
+
+# DEB-specific options
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Unicode Consortium <grhoten@gmail.com>")
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "libicu-dev (>= 77.1)")
```
grhoten, line 144: The dev part is only needed for building. The non-dev dependency is needed for runtime.

.github/workflows/create-ubuntu-distribution-package.yml
```diff
@@ -0,0 +1,81 @@
+name: Build & Package (Ubuntu)
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+  pull_request:
+    branches: [ main, master ]
```
nciric, line 9: I don't think we need the master part. It's obsolete and our repo doesn't have it.

.github/workflows/create-ubuntu-distribution-package.yml
```diff
@@ -0,0 +1,81 @@
...
+  build-and-package:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        lfs: true
+
+    - name: Setup Git LFS
```
nciric, line 21: We can land it as is for now, but fix it after Frank lands 128 (LFS caching).

.github/workflows/create-ubuntu-distribution-package.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: |
+        git lfs install
+        git lfs pull
+
+    - name: Install system dependencies
+      run: |
+        sudo apt update
+        sudo apt install -y cmake build-essential clang pkg-config
+
+    - name: Install ICU 77.1
```
nciric, line 31: ICU caching already landed - we can use it here to save bandwidth.

inflection/CMakeLists.txt
nciric: Is that a mandatory field? If we have to have it, we can use the inflection-team@unicode.org one.

nciric (commented):
Looks good in general with some comments.

inflection/CMakeLists.txt
nciric: Agree.

.github/workflows/create-ubuntu-distribution-package.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: |
+        cd /tmp
+        wget https://github.com/unicode-org/icu/releases/download/release-77-1/icu4c-77_1-src.tgz
```
nciric, line 34: Why do we download and compile sources? Why not just download and install binary release for Ubuntu? It takes time to compile ICU.

One needs to only copy the $download_location/usr tree to /usr/local/ folder

nciric: See https://github.com/unicode-org/inflection/pull/128/files for example on how to do it.

.github/workflows/create-ubuntu-distribution-packaging.yml
```diff
@@ -0,0 +1,98 @@
+name: Build & Package (Ubuntu)
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+  pull_request:
```
nciric, line 8: Would this run the flow on each pull request? I don't think we need that.

.github/workflows/create-ubuntu-distribution-packaging.yml
```diff
@@ -0,0 +1,98 @@
...
+          tar -xzf icu4c-77_1-Ubuntu22.04-x64.tgz -C icu-install
+          sudo cp -r icu-install/icu/usr/local/* /usr/local/
+          sudo ldconfig
+
+      - name: Setup ICU (from cache)
+        if: steps.cache-icu.outputs.cache-hit == 'true'
+        run: |
+          sudo ldconfig
+
+      - name: Debug ICU version
```
nciric, line 52: Do we need this for final version?

.github/workflows/create-ubuntu-distribution-packaging.yml
```diff
@@ -0,0 +1,98 @@
...
+          CC=clang CXX=clang++ cmake .. \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DICU_ROOT=/usr/local \
+            -DCMAKE_PREFIX_PATH=/usr/local
+          make -j$(nproc)
+
+      - name: Run tests
+        run: |
+          cd inflection/build
+          make check
```
nciric, line 72: I think you can run test in parallel as with previous step (-j$...)

nciric (changes requested):

grhoten:
Is it possible to derive the version of the Debian package from resources/share/inflection/config.properties?

.github/workflows/create-ubuntu-distribution-packaging.yml
dockerh: I’ll remove the pull_request trigger so the workflow runs only on tag pushes and manual dispatch.

.github/workflows/create-ubuntu-distribution-packaging.yml
dockerh: This was helpful for initial testing, but I agree,  it can be removed from the final version. I’ll delete the debug step.

.github/workflows/create-ubuntu-distribution-packaging.yml
dockerh: Yes, 
I’ll update the make check step to use -j$(nproc) for parallel test execution.

.github/workflows/create-ubuntu-distribution-packaging.yml
```diff
@@ -0,0 +1,98 @@
...
+        id: cache-icu
+        with:
+          path: /usr/local
+          key: icu-77.1-ubuntu-${{ runner.os }}
+
+      - name: Install ICU 77.1 (Binary)
+        if: steps.cache-icu.outputs.cache-hit != 'true'
+        run: |
+          cd /tmp
+          wget https://github.com/unicode-org/icu/releases/download/release-77-1/icu4c-77_1-Ubuntu22.04-x64.tgz
```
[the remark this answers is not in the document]
dockerh: Let me look into it, Currently, this workflow targets x64. I’ll explore adding runs-on: ubuntu-latest matrix support for arm64.

dockerh:
> Is it possible to derive the version of the Debian package from resources/share/inflection/config.properties?

Great suggestion, I'll look into it.

inflection/CMakeLists.txt
```diff
@@ -130,3 +130,51 @@ make generate-coverage-csv  : Generates code coverage as a csv\\n\
...
+  list(GET INFLECTION_VERSION_LIST 0 CPACK_PACKAGE_VERSION_MAJOR)
+endif()
+if(_len GREATER 1)
+  list(GET INFLECTION_VERSION_LIST 1 CPACK_PACKAGE_VERSION_MINOR)
+endif()
+if(_len GREATER 2)
+  list(GET INFLECTION_VERSION_LIST 2 CPACK_PACKAGE_VERSION_PATCH)
+endif()
+
+
```
nciric, line 160: remove extra new lines (line 141 and here).

inflection/CMakeLists.txt
```diff
@@ -130,3 +130,51 @@ make generate-coverage-csv  : Generates code coverage as a csv\\n\
...
 install(DIRECTORY ${INFLECTION_DATA_ROOT}/    TYPE DATA    COMPONENT inflection_data)
+
+
+
+# CPack Configuration for Ubuntu Packaging
+
+set(CPACK_PACKAGE_NAME "unicode-inflection")
+set(CPACK_PACKAGE_VERSION "${INFLECTION_VERSION}")
+
+
```
nciric, line 141: I would put a comment here:
# Apply the current tagged Inflection version to the CPack version.
# CPack inherits version from ??? so we need to reset before assignment.

.github/workflows/create-ubuntu-distribution-packaging.yml
```diff
@@ -0,0 +1,94 @@
+name: Build & Package (Ubuntu)
```
nciric, line 1: We should have a comment on top (or a separate document) that tells us how to update ICU dependancy:

# To update ICU version:
# 1. Update ICU_VERSION variable
# 2. Check package names, e.g. is it still Ubuntu22.04-
# 3. Update inflection/CMakeLists.txt where we check CPACK_DEBIAN_PACKAGE_DEPENDS "libicu77 (>= 77.1)"

Anything else I missed?

nciric (approved):

inflection/CMakeLists.txt
dockerh: Okay

.github/workflows/create-ubuntu-distribution-packaging.yml
dockerh: Good idea, will do that.

nciric (approved):

Issue #134, opened 2025-06-17
use std::unique_ptr


FrankYFTang:
Some of the code get raw pointer and then delete, we should switch to std::unique_ptr to avoid memory leak

PR #142, opened 2025-06-27
Inflection 134 std::unique_ptr for DictionaryKeyIterator.cpp

FrankYFTang:
#134


inflection/src/inflection/dictionary/DictionaryKeyIterator.cpp
```diff
@@ -18,24 +18,22 @@ DictionaryKeyIterator::DictionaryKeyIterator(const metadata::MarisaTrieIterator<
 }
 
 DictionaryKeyIterator::DictionaryKeyIterator(DictionaryKeyIterator&& iterator)
-    : trieIterator(iterator.trieIterator)
+    : trieIterator(iterator.trieIterator.release())
```
grhoten, line 21: If you're removing the npc checks elsewhere, then this line needs an null pointer check added here.

grhoten (approved):

PR #143, opened 2025-06-27
Inflection 134 std::unique_ptr for DictionaryMetaData.cpp

FrankYFTang:
#134


inflection/src/inflection/dictionary/DictionaryMetaData.cpp
```diff
@@ -116,12 +115,12 @@ ::std::u16string* DictionaryMetaData::transform(::std::u16string* dest, std::u16
 int64_t* DictionaryMetaData::getCombinedBinaryType(int64_t* result, std::u16string_view word) const
 {
     *npc(result) = 0;
-    auto combinedType = npc(dictionary)->getWordType(word);
+    auto combinedType = dictionary->getWordType(word);
```
grhoten, line 118: If you're going to remove the null pointer checks like this, then the check needs to move to the constructor when the dictionary is assigned. It should never be null. Once the constructor has the null pointer check, I'll approve this pull request.

inflection/src/inflection/dictionary/Inflector.cpp
```diff
@@ -13,7 +13,7 @@
 namespace inflection::dictionary {
 
 const Inflector& Inflector::getInflector(const ::inflection::util::ULocale &locale) {
-    const Inflector* result = npc(npc(DictionaryMetaData::createDictionary(locale))->dictionary)->inflector;
+    const Inflector* result = (npc(DictionaryMetaData::createDictionary(locale))->dictionary)->inflector;
     if (result == nullptr) {
         throw ::inflection::exception::NullPointerException(inflection::util::StringViewUtils::to_u16string("Inflector not found for " + locale.getName()));
     }
```
grhoten, line 19: The extra parentheses can be removed.

It's worth noting that the code does try its best to create a stub dictionary, but this null check for the inflector result does trigger when the **required** dictionary was not found. I think the revision is OK, but we have to be extra sensitive here to fail safely instead of crash by derefrencing null.

grhoten (changes requested):

inflection/src/inflection/dictionary/Inflector.cpp
FrankYFTang: PTAL

inflection/src/inflection/dictionary/DictionaryMetaData.cpp
FrankYFTang: PTAL

grhoten (approved):

Issue #147, opened 2025-06-30
Need to fix warning generated by doxygen


FrankYFTang:
doxygen generate the following warnings
in https://github.com/unicode-org/inflection/actions/runs/15939468597/job/44965330600?pr=146

We should fix them

```
/home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/dialog/SemanticValue.hpp:55: warning: no uniquely matching class member found for 
  const ::std::u16string & inflection::dialog::SemanticValue::getValue() const
Possible candidates:
  'const ::std::u16string & inflection::getValue() const' at line 26 of file /home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/dialog/SemanticValue.hpp
  'virtual std::u16string inflection::getValue() const' at line 33 of file /home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/lang/features/LanguageGrammarFeatures_Feature.hpp
/home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/dialog/SemanticValue.hpp:61: warning: no matching class member found for 
  const ::std::u16string & inflection::dialog::SemanticValue::getLabel() const
Possible candidates:
  'const ::std::u16string & inflection::getLabel() const' at line 31 of file /home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/dialog/SemanticValue.hpp
/home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/exception/ICUException.hpp:63: warning: no uniquely matching class member found for 
Possible candidates:
  'bool inflection::operator>(const ::inflection::util::ULocale &other) const' at line 165 of file /home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/util/ULocale.hpp
/home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/util/ULocale.hpp:254: warning: no matching class member found for 
  std::string_view inflection::util::ULocale::getCountry() const
Possible candidates:
  'std::string_view inflection::getCountry() const' at line 185 of file /home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/util/ULocale.hpp
/home/runner/work/inflection/inflection/inflection/build/inflection_headers/inflection/util/ULocale.hpp:260: warning: no uniquely matching class member found for 
  std::string_view inflection::util::ULocale::getLanguage() const
Possible candidates:
[...]

PR #148, opened 2025-07-01
Change namespace in message2 (issue/147)

FrankYFTang:
#147 
Restructure comment per @grhoten

### DIFF
--- a/inflection/src/inflection/message2/MF2Factory.hpp
+++ b/inflection/src/inflection/message2/MF2Factory.hpp
@@ -3,7 +3,7 @@
  */
 #pragma once
 
-#include <inflection/api.h>
+#include <inflection/message2/fwd.hpp>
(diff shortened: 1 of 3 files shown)


inflection/src/inflection/message2/MF2Factory.hpp
```diff
@@ -13,19 +13,18 @@ class FormatterFactory;
 class SelectorFactory;
 }  // namespace U_ICU_NAMESPACE::message2
```
grhoten, line 14: Typically these internal declarations go to the forward header too.

FrankYFTang: SelectorFactory and FormatterFactory are not internal, but external

FrankYFTang:
PTAL.

PR #151, opened 2025-07-03
Fix some warning in doxygen (issue 147)

FrankYFTang:


grhoten (approved):
I should probably trim some of those unused LocaleUtils locales, but that’s not the issue here. These changes look good.

PR #152, opened 2025-07-03
Add comment for super class documentation Fix 147 

FrankYFTang:
#147


FrankYFTang:
@grhoten I am not sure this is the right fix. But if we do not document it doxygen will consider it missing info and give us warning

FrankYFTang:
> Would using `TYPEDEF_HIDES_STRUCT = YES` help? That may require the least maintenance.

no. Warning still there. I do not think it is a struct here.

FrankYFTang:
PTAL

grhoten (approved):
Works for me.

grhoten:
Thanks for the change.

PR #160, opened 2025-07-09
Add operator<=> and remove operator< operator and compareTo function (issue 155)

FrankYFTang:
Address #155 

Also correct the comparsion of LocaleUtil and the test which depend on reverse behavior.


FrankYFTang:
@grhoten any opinion about this?

FrankYFTang:
@nciric @markusicu and I discussed this in our weekly meeting and we dig into one of our portability document and find we may have problem to land this change due to the C++20 requirement. I would like to abandon this efforts but take some issue I learn from this PR to a different PR to address some issues I discovered and fix in this PR.

FrankYFTang:
Park this PR until we all feel comfortable to require C++20

FrankYFTang:
have some 2nd thought. reopen it. we should face to face discuss this first.

PR #162, opened 2025-07-15
Inflection 134 Use std::unique_ptr to avoid delete

FrankYFTang:
#134

### DIFF
--- a/inflection/src/inflection/dictionary/Inflector.hpp
+++ b/inflection/src/inflection/dictionary/Inflector.hpp
@@ -41,6 +41,7 @@ private:
     const Inflector_MMappedDictionary mmappedDictionary;
 
     friend class DictionaryMetaData_MMappedDictionary;
+    friend struct std::default_delete<Inflector>;
     friend class Inflector_Inflection;
     friend class Inflector_InflectionPattern;
 };
--- a/inflection/src/inflection/dictionary/Inflector.hpp
+++ b/inflection/src/inflection/dictionary/Inflector.hpp
@@ -33,6 +33,7 @@ public:
 private:
     explicit Inflector(inflection::util::MemoryMappedFile& memoryMappedFile, const ::std::u16string& sourcePath, const ::inflection::dictionary::DictionaryMetaData_MMappedDictionary &dictionary);
 
+public:
     ~Inflector() override;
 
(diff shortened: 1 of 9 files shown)


inflection/src/inflection/dictionary/Inflector.hpp
```diff
@@ -41,6 +41,7 @@ class inflection::dictionary::Inflector
     const Inflector_MMappedDictionary mmappedDictionary;
 
     friend class DictionaryMetaData_MMappedDictionary;
+    friend struct std::default_delete<Inflector>;
```
grhoten, line 44: This looks suspicious. What's going on here?

inflection/src/inflection/tokenizer/TokenizerFactory.cpp
```diff
@@ -146,7 +146,7 @@ Tokenizer* TokenizerFactory::createTokenizerObject(const util::ULocale& locale,
                 }
                 auto tokenizer = new Tokenizer(TOKENIZER_CONSTRUCTORS[mid].construct(locale, systemConfig));
                 // Initialize the data loading and caches so that it doesn't possibly fail later.
-                delete npc(tokenizer)->createTokenChain(u"initialization");
+                ::std::unique_ptr<TokenChain> unused(npc(tokenizer)->createTokenChain(u"initialization"));
```
grhoten, line 149: The original was actually correct. There is no intermediate line between getting the object and deleting it. I recommend leaving this one alone.

FrankYFTang:
PTAL

inflection/src/inflection/dictionary/Inflector.hpp
grhoten: Such a declaration is missing the point of making it private in the first place. It's to prevent accidental deletion by non-friends. If std::unique_ptr is to be used to delete it, then it's better to just make the destructor public. It is odd to make a subclass destructor like this, and it's better to just make it public than to add another friend.

grhoten:
I think the only remaining change needed is the Inflector.hpp change to make the destructor public is needed.

FrankYFTang:
Changed according to comments. PTAL

grhoten (approved):

PR #164, opened 2025-07-15
issue 163 fix ULocale < and > operator order

FrankYFTang:
Also fix incorrect test and enhance it.


grhoten (approved):

grhoten:
After some thought, I think it was done this way to get the tests to run in the right order. Probably std::less has to be used in such situations.

PR #169, opened 2025-07-29
Multi-Platform GitHub Actions Workflow for Inflection Library

dockerh:
Adds a GitHub Actions workflow to build, test, and package the Unicode Inflection library for Ubuntu and macOS.


.github/workflows/cmake-multi-platform.yml
```diff
@@ -60,96 +51,112 @@ jobs:
...
+            brew install git-lfs
+          fi
+          git lfs install
+
+      - name: Pull LFS files
+        run: git lfs pull
+        working-directory: ${{ github.workspace }}
+
+      - name: Debug Data Directory
+        run: ls -R inflection/resources/org/unicode/inflection/dictionary || true
```
grhoten, line 83: I think you meant `-o`, which shows the sizes, and dates, which are important. The `-R` lists the directory recursively.

grhoten (changes requested):
The macOS build seems to be failing because it's encountering some sort of issue with Debian packages.

It seems that these changes fixes 1 of 2 Ubuntu builds. That's an improvement. Can you fix both of them? It looks like one of them doesn't get the right LFS content.

Also before merging, please remove the content from my pull request.  I'd like to keep the pull requests separate for tracking purposes.

grhoten:
It seems that #170 fixed the underlying issue. Thanks for helping to debug this.

Issue #172, opened 2025-08-12
Expand Serbian support with rule based inflection


nciric:
The Serbian language has four classes of noun declension.  Implementing them significantly reduces the need for a large dictionary.

The only nouns needing a dictionary are exceptions to the rules.

Here's the [link](https://pspasojevic.blogspot.com/2013/04/blog-post_1622.html) to the rules (in Serbian).

PR #173, opened 2025-08-12
Implementing group 3 noun rules for Serbian.

nciric:
Fully implements group 3 (all nouns ending with -a).

This removes a need for a large number of nouns to be added to Wikidata.

Resolves part of #172 .

### DIFF
--- a/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
+++ b/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
@@ -139,7 +139,7 @@ bool isProperNoun(const ::std::u16string &lemma);
     return inflection;
 }
 
-::inflection::dialog::DisplayValue *SrGrammarSynthesizer_SrDisplayFunction::getDisplayValue(const dialog::SemanticFeatureModel_DisplayData &displayData, const ::std::map<::inflection::dialog::SemanticFeature, ::std::u16string> &constraints, bool /* enableInflectionGuess */) const
+::inflection::dialog::DisplayValue *SrGrammarSynthesizer_SrDisplayFunction::getDisplayValue(const dialog::SemanticFeatureModel_DisplayData &displayData, const ::std::map<::inflection::dialog::SemanticFeature, ::std::u16string> &constraints, bool enableInflectionGuess) const
 {
     ::std::u16string displayString;
     if (!displayData.getValues().empty()) {
@@ -150,7 +150,7 @@ bool isProperNoun(const ::std::u16string &lemma);
     }
     if (dictionary.isKnownWord(displayString)) {
         displayString = inflectFromDictionary(constraints, displayString);
-    } else {
+    } else if (enableInflectionGuess) {
         // Let's use rule based inflection for nouns. Assume lemma is singular, nominative.
         displayString = inflectWithRule(constraints, displayString);
     }
--- a/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
+++ b/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
@@ -199,23 +199,15 @@ Syllables countSyllables(const ::std::u16string& lemma) {
     }
 }
 
-::std::u16string inflectByRuleOE(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase, const ::std::u16string &gender)
+::std::u16string inflectByRuleOE(const ::std::u16string &lemma, [[maybe_unused]] const ::std::u16string &number, [[maybe_unused]] const ::std::u16string &targetCase, [[maybe_unused]] const ::std::u16string &gender)
 {
     // TODO(nciric): implement logic.
-    // Silence unused variable warnings
-    auto base = number;
-    base = targetCase;
-    base = gender;
     return lemma;
 }
 
-::std::u16string inflectByRuleE(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase, const ::std::u16string &gender)
+::std::u16string inflectByRuleE(const ::std::u16string &lemma, [[maybe_unused]] const ::std::u16string &number, [[maybe_unused]] const ::std::u16string &targetCase, [[maybe_unused]] const ::std::u16string &gender)
 {
     // TODO(nciric): implement logic.
-    // Silence unused variable warnings
-    auto base = number;
-    base = targetCase;
-    base = gender;
     return lemma;
 }
 
@@ -255,13 +247,9 @@ Syllables countSyllables(const ::std::u16string& lemma) {
     return base;
 }
 
-::std::u16string inflectByRuleConsonant(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase, const ::std::u16string &gender)
+::std::u16string inflectByRuleConsonant(const ::std::u16string &lemma, [[maybe_unused]] const ::std::u16string &number, [[maybe_unused]] const ::std::u16string &targetCase, [[maybe_unused]] const ::std::u16string & gender)
 {
     // TODO(nciric): implement logic.
-    // Silence unused variable warnings
-    auto base = number;
-    base = targetCase;
-    base = gender;
     return lemma;
 }
 
--- a/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
+++ b/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
@@ -213,8 +213,9 @@ Syllables countSyllables(const ::std::u16string& lemma) {
 
 ::std::u16string inflectByRuleA(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase)
 {
-    static constexpr auto suffix_sg = ::std::to_array<::std::u16string>({u"а", u"е", u"и", u"у", u"а", u"ом", u"и"});
-    static constexpr auto suffix_pl = ::std::to_array<::std::u16string>({u"е", u"а", u"ама", u"е", u"е", u"ама", u"ама"});
+    // 2025-08: MacOS doesn't like to_array, so we'll initialize manually.
+    static constexpr ::std::array<::std::u16string, NUMBER_OF_CASES> suffix_sg = {u"а", u"е", u"и", u"у", u"а", u"ом", u"и"};
+    static constexpr ::std::array<::std::u16string, NUMBER_OF_CASES> suffix_pl = {u"е", u"а", u"ама", u"е", u"е", u"ама", u"ама"};
 
     ::std::u16string base = lemma;
     // Remove trailing a and apply suffix.
--- a/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
+++ b/inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
@@ -11,10 +11,12 @@
 #include <inflection/dialog/SemanticFeatureModel.hpp>
 #include <inflection/dialog/SemanticFeatureModel_DisplayData.hpp>
 #include <inflection/dialog/DisplayValue.hpp>
+#include <inflection/dictionary/PhraseProperties.hpp>
 #include <inflection/grammar/synthesis/GrammemeConstants.hpp>
 #include <inflection/grammar/synthesis/GrammarSynthesizerUtil.hpp>
 #include <inflection/lang/StringFilterUtil.hpp>
 #include <inflection/util/LocaleUtils.hpp>
+#include <inflection/util/StringViewUtils.hpp>
 #include <inflection/util/UnicodeSetUtils.hpp>
 #include <inflection/npc.hpp>
 #include <array>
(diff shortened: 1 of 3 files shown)


inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
...
+    return base;
+}
+
+::std::u16string inflectByRuleConsonant(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase, const ::std::u16string &gender)
+{
+    // TODO(nciric): implement logic.
+    // Silence unused variable warnings
+    auto base = number;
+    base = targetCase;
+    base = gender;
```
grhoten: This is normally silenced by commenting out the argument instead of trying to use it.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
...
+
+::std::u16string inflectByRuleA(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase)
+{
+    static constexpr auto suffix_sg = ::std::to_array<::std::u16string_view>({u"а", u"е", u"и", u"у", u"а", u"ом", u"и"});
+    static constexpr auto suffix_pl = ::std::to_array<::std::u16string_view>({u"е", u"а", u"ама", u"е", u"е", u"ама", u"ама"});
+
+    ::std::u16string base = lemma;
+    // Remove trailing a and apply suffix.
+    base.pop_back();
+    base = applySuffix(base, suffix_sg, suffix_pl, number, targetCase);
```
grhoten, line 227: For this kind of mapping, you may be inspired by Arabic, German or Italian. They convert a string to a numeric key (makeLookupKey) containing multiple grammemes, and they map the key to a string. This mapping is initialized in the constructor instead of at runtime.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
...
+    return lemma;
+}
+
+::std::u16string inflectByRuleE(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase, const ::std::u16string &gender)
+{
+    // TODO(nciric): implement logic.
+    // Silence unused variable warnings
+    auto base = number;
+    base = targetCase;
+    base = gender;
```
grhoten: This is normally silenced by commenting out the argument instead of trying to use it.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
...
+    }
+}
+
+::std::u16string inflectByRuleOE(const ::std::u16string &lemma, const ::std::u16string &number, const ::std::u16string &targetCase, const ::std::u16string &gender)
+{
+    // TODO(nciric): implement logic.
+    // Silence unused variable warnings
+    auto base = number;
+    base = targetCase;
+    base = gender;
```
grhoten: This is normally silenced by commenting out the argument instead of trying to use it.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
...
+    }
+
+    if (total == 1) {
+        return Syllables::ONE_SYLLABLE;
+    } else if (total == 2) {
+        return Syllables::TWO_SYLLABLES;
+    } else {
+        return Syllables::MULTI_SYLLABLES;
+    }
+}
```
grhoten: What do you think of this?
```
static bool isConsonant(char16_t ch) {
    return morphun::lang::StringFilterUtil::CYRILLIC_SCRIPT().contains(ch) && !morphun::dictionary::PhraseProperties::DEFAULT_VOWELS_START().contains(ch);
}

enum class Syllables {
    ONE_SYLLABLE,
    TWO_SYLLABLES,
    MULTI_SYLLABLES,
};

Syllables countSyllables(const ::std::u16string& word) {
    int32_t total = 0;
    if (word.length() > 2) {
        std::u16string lowercase;
        morphun::util::StringUtils::lowercase(&lowercase, word, ::morphun::util::LocaleUtils::SERBIAN());
        size_t startIndex = 0;
        size_t nextR;
        const auto stringLengthWithVowelSuffix = lowercase.length() - 1; // Always enough room for a suffix letter
        while ((nextR = lowercase.find(u'р', startIndex)) != std::string::npos && nextR < stringLengthWithVowelSuffix) {
            ++total; // +1 for r
            if (isConsonant(lowercase[nextR + 1]) && (nextR == 0 || isConsonant(lowercase[nextR - 1]))) {
                ++total;
            }
            startIndex = nextR + 1;
        }
    }

    if (total == 1) {
        return Syllables::ONE_SYLLABLE;
    } else if (total == 2) {
        return Syllables::TWO_SYLLABLES;
    } else {
        return Syllables::MULTI_SYLLABLES;
    }
}
```

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -17,8 +17,12 @@
 #include <inflection/util/LocaleUtils.hpp>
 #include <inflection/util/UnicodeSetUtils.hpp>
 #include <inflection/npc.hpp>
+#include <icu4cxx/RegularExpression.hpp>
+#include <array>
 #include <iterator>
 #include <memory>
+#include <string>
+#include "SrGrammarSynthesizer_SrDisplayFunction.hpp"
```
grhoten: Hopefully that path always works. Sometimes that can be iffy.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
...
+    if (number == GrammemeConstants::NUMBER_PLURAL() && targetCase == GrammemeConstants::CASE_GENITIVE()) {
+        if (lemma.ends_with(u"тња") || lemma.ends_with(u"дња") || lemma.ends_with(u"пта") || lemma.ends_with(u"лба") || lemma.ends_with(u"рва")) {
+            base.back() = u'и';
+        }
+        size_t pos = 0;
+        if ((pos = base.rfind(u"јк")) != ::std::u16string::npos) base.replace(pos, 2, u"јак");
+        if ((pos = base.rfind(u"мљ")) != ::std::u16string::npos) base.replace(pos, 2, u"маљ");
+        if ((pos = base.rfind(u"вц")) != ::std::u16string::npos) base.replace(pos, 2, u"вац");
+        if ((pos = base.rfind(u"тк")) != ::std::u16string::npos) base.replace(pos, 2, u"так");
+        if ((pos = base.rfind(u"пк")) != ::std::u16string::npos) base.replace(pos, 2, u"пак");
```
grhoten: Is this supposed to be a suffix match, or should it match prefixes and infixes?

What do you think of this code? This code assumes that more string replacements may be done in the future with less copy and paste code. If you really meant ends_with instead of rfind, then it's easy to insert a break on the first match.
```
    size_t pos = 0;
    static const char16_t * mappings[][2] = {
        {u"јк", u"јак"},
        {u"мљ", u"маљ"},
        {u"вц", u"вац"},
        {u"тк", u"так"},
        {u"пк", u"пак"},
    };
    for (const auto &[str, replacement] : mappings) {
        if ((pos = base.rfind(str)) != ::std::u16string::npos) {
            base.replace(pos, std::u16string_view(str).length(), replacement);
        }
    }
```

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -17,8 +17,12 @@
 #include <inflection/util/LocaleUtils.hpp>
 #include <inflection/util/UnicodeSetUtils.hpp>
 #include <inflection/npc.hpp>
+#include <icu4cxx/RegularExpression.hpp>
```
grhoten: It looks like this line isn't needed anymore.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: I used [[maybe_unused]]  instead of comments.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: [[maybe_unused]]

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: Done

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: I don't know how this line was added (maybe auto complete/CoPilot). Removed it as I already have it with a full path on top.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: [[maybe unused]] instead.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: it was a suffix match - I tweaked the code to use ends_with as that reflect the actual intent.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: Your code skips vowels (only counts Rs). I like the isConsonant, and I added isVowel function and used it in the code.
It looks simpler now.

PTAL

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: Is the concern the runtime size increase (static constexpr)? If yes, I can remove the static (creating these arrays is cheap).
Otherwise the current approach looks simpler. I will look into refactoring this code as I add more cases, potentially implementing Arabic like approach.

WDYT?

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -87,9 +149,154 @@ ::inflection::dialog::DisplayValue * SrGrammarSynthesizer_SrDisplayFunction::get
         return nullptr;
     }
     if (dictionary.isKnownWord(displayString)) {
-        displayString = inflectString(constraints, displayString);
+        displayString = inflectFromDictionary(constraints, displayString);
+    } else {
```
grhoten: In this scenario, you should check `enableInflectionGuess` before using the rules.

inflection/test/resources/inflection/dialog/inflection/sr.xml
```diff
@@ -18,4 +18,25 @@
...
+    <test><source case="vocative" number="singular" gender="masculine" pos="noun">Љуба</source><result>Љубо</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">пратња</source><result>пратњи</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">радња</source><result>радњи</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">лопта</source><result>лопти</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">молба</source><result>молби</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">конзерва</source><result>конзерви</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">гошћа</source><result>гошћа</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">двојка</source><result>двојака</result></test>
+    <test><source case="genitive" number="plural" gender="feminine" pos="noun">битка</source><result>битака</result></test>
```
grhoten, line 39: You have a lot of fully fleshed out constraints. Most of the other languages only change specific grammemes. Sometimes you only specify the case, number or gender. The other tests usually specify less. The other languages usually default to noun. These tests are currently fine, but common usage starts from any surface form (ideally a unique surface form), and then you modify just the relevant grammemes.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
grhoten: It was to make it more scalable, but this is fine too.

grhoten (approved):
Changes look fine. Optional comments to consider where also provided.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: done, added else if (enableInflectionGuess) {...

inflection/test/resources/inflection/dialog/inflection/sr.xml
nciric: I can probably remove noun info, but I do need a case, gender and number for rule based approach to work.
I also assume nominative input (lemma) - otherwise the rules would be more complex, or would need dictionary support to implement them.

Issue #77, opened 2025-02-04
Add API for stating the source grammemes instead of using a SemanticConcept


grhoten:
Right now, a lot of the code assumes that a given word is a lemma or makes assumptions about the grammemes of an ambiguous word. The current API just have a way to add constraints for adding the desired grammemes. We should consider adding API for adding such information to disambiguate the starting point in an inflection table.

For example, a given word may be masculine or feminine.  When a word has 2 different meanings that depends on the gender, it can be hard to determine the right grammatical gender to use for the definite article.  Such information has usually been disambiguated through a SemanticConcept, but there may be times that you want to use it in the current message without a SemanticConcept.

One possible path to implement such a change is to change InflectableStringConcept to provide some SemanticConcept functionality for stating the source grammemes for inflection. It would add some data to the DisplayValue.  The various grammar synthesizers would need to start reading that data little more consistently though.

anushka-cseatmnc:
Hi @grhoten ,I’ve been going through the issue [#77 Add API for stating the source grammemes instead of using a SemanticConcept](https://github.com/unicode-org/inflection/issues/77), and I believe I can contribute to it. I would love to work on designing and implementing the API for adding source grammemes to disambiguate inflection tables.

Would it be possible for you to assign this issue to me? I’m excited to start working on it and learn more about the inflection system.

Thank you!

Best regards,
Anushka

grhoten:
Thanks for your interest. This would actually be a sizable change, and it's a nuanced feature enhancement. It would not be a good task for someone new to this project. There are easier tasks to work on that have a more immediate impact on getting the 0.1 release out the door, like getting the lexical dictionaries transitioned to using Wikidata as the source, or writing a script to automate such a task.

With that being said, here's why this issue is hard.

Right now, the constraints provide some ways to state which inflection pattern is desired. You can specify in the constraints that you want the noun inflection or the verb inflection. Take "[brush](https://en.wiktionary.org/wiki/brush#English)" as example. The current constraint model allows you to differentiate between "1 brush/2 brushes" by interpreting it as a noun, and you can specify a verb for "she brushes/he brushes/it brushes/they brush".

[...]

PR #174, opened 2025-08-12
Inflection 77: Adding C and C++ API for stating source grammemes and tests for the same.

BHK4321:
@grhoten 

Issue #77 
Could you please review this so that I could move forward for C API and XML parsing too.


inflection/src/inflection/dialog/InflectableStringConcept.hpp
```diff
@@ -69,6 +69,15 @@ class INFLECTION_CLASS_API inflection::dialog::InflectableStringConcept
...
     InflectableStringConcept(const SemanticFeatureModel* model, const SpeakableString& value);
+
+    /**
+     * Constructs a concept given a semantic feature model and a speakable string
+     *
+     * @param model - The semantic feature model required to initialize the concept.
+     * @param value - The speakable string to convert to a concept
+     * @param intitialConstraints - The intitial constraints for the map.
+     */
+    InflectableStringConcept(const SemanticFeatureModel* model, const SpeakableString& value, const std::map<SemanticFeature, ::std::u16string> intitialConstraints);
```
grhoten: The map should be a `const &` for the API.

inflection/src/inflection/dialog/DisplayValue.hpp
```diff
@@ -72,6 +72,13 @@ class INFLECTION_CLASS_API inflection::dialog::DisplayValue
      * @param speakFeature The speakFeature from the SemanticFeatureModel that represents the SemanticFeature for the speak information for a SpeakableString.
      */
     DisplayValue(const SpeakableString& value, const SemanticFeature& speakFeature);
+    /**
+     * Construct a display value with a SpeakableString.
+     * @param value A SpeakableString
+     * @param speakFeature The speakFeature from the SemanticFeatureModel that represents the SemanticFeature for the speak information for a SpeakableString.
+     * @param constraintMap The intitial constraint map.
+     */
+    DisplayValue(const SpeakableString& value, const SemanticFeature& speakFeature, const ::std::map<SemanticFeature, ::std::u16string> constraintMap);
```
grhoten: The map should be a `const &` for the API.

inflection/src/inflection/dialog/DisplayValue.cpp
```diff
@@ -21,14 +24,26 @@ DisplayValue::DisplayValue(const ::std::u16string& value)
...
 
+DisplayValue::DisplayValue(
+    const SpeakableString& value,
+    const SemanticFeature& speakFeature)
+    : DisplayValue(value.getPrint(), {})
+{
+    if (!value.speakEqualsPrint()) {
+        this->constraintMap.emplace(speakFeature, value.getSpeak());
+    }
+}
```
grhoten, line 45: Can you please declare these functions in the same order as the header? It makes it easier to navigate.

Also the 2 argument constructor with the speakFeature can now call the new constructor with an empty map. Then the redundant if statement can be removed. It's best to keep the copy and paste code to a minimum to improve the maintenance.

inflection/src/inflection/dialog/InflectableStringConcept.cpp
```diff
@@ -15,7 +15,21 @@
...
+    : super(model)
+    , value(value)
+    , defaultDisplayValue(value, *npc(super::getSpeakFeature()), intitialConstraints)
+{
+}
+
+InflectableStringConcept::InflectableStringConcept(
+    const SemanticFeatureModel* model,
+    const SpeakableString& value
+)
```
grhoten: Can you please declare these functions in the same order as the header? It makes it easier to navigate.

Also the 2 argument constructor with the SpeakableString can now call the new constructor with an empty map. Then the copied code can be removed. It's best to keep the copy and paste code to a minimum to improve the maintenance.

BHK4321: Sure I would keep that in mind.
Being a fresher I wasn't aware.

BHK4321:
@grhoten I have added the C API too can you please review and tell me the next steps.

inflection/src/inflection/dialog/InflectableStringConcept.h
```diff
@@ -34,6 +34,20 @@ INFLECTION_CAPI IDInflectableStringConcept* iinf_toInflectableStringConcept(IDSe
...
+ * @param value - The speakable string to convert to a concept
+ * @param features - The initial semantic features to apply to the concept
+ * @param values - The initial corresponding values for the semantic features
+ * @param SemanticFeaturesLen - The number of semantic features and values provided
+ * @param status Must be a valid pointer to an error code value,
+ *        which must not indicate a failure before the function call.
+ *        This is set to a failure when a failure has occurred during execution.
+ */
+INFLECTION_CAPI IDInflectableStringConcept* iinf_createWithConstraints(const IDSemanticFeatureModel* model, const IDSpeakableString* value, 
+                                                                       const inflection::dialog::SemanticFeature* features, const char16_t** values, int32_t SemanticFeaturesLen, UErrorCode* status);
```
grhoten: C++ API is not allowed in C API. Please see `ipron_createWithDefaults` for how to pass in such values. That API also ensures that the 2 lengths are exactly the same.

For consistency with the PronounConcept API, it would be ideal if this was also named `iinf_createWithDefaults` to be consistent with the naming.

BHK4321:
@grhoten 
I have updated the code accordingly, could you please consider merging this PR I will continue with this issue #77 (the testing and XML parsing) after some time.
Thanks

inflection/src/inflection/dialog/InflectableStringConcept.h
```diff
@@ -6,7 +6,7 @@
 #include <inflection/dialog/SpeakableString.h>
 #include <inflection/dialog/SemanticFeatureConcept.h>
 #include <inflection/dialog/SemanticFeatureModel.h>
-
+#include <inflection/dialog/SemanticUtils.hpp>
```
grhoten: This internal C++ header should not be used in public C API.

inflection/src/inflection/dialog/PronounConcept.cpp
```diff
@@ -211,6 +211,7 @@ PronounConcept::PronounConcept(const SemanticFeatureModel& model, const ::std::v
     if (pronounData->numValues() == 0) {
         throw ::inflection::exception::IllegalArgumentException(u"Display data can not be empty.");
     }
+
```
grhoten: I don't think that this file needs to change.

grhoten (changes requested):
All new API need tests. The code should always be in a releasable state. Please add tests for these changes. The current code should be well over 90% line code coverage.

Please also state which issue is being fixed with these changes. That is best done with prefixing your commits and pull request summary with "Inflection-77".

inflection/src/inflection/dialog/PronounConcept.cpp
BHK4321: Yup sure, I mistakenly added a line and forgot

inflection/src/inflection/dialog/InflectableStringConcept.h
BHK4321: I will fix that

BHK4321:
@grhoten 
Please review this commit and I had some questions while working too:

- Is the `compareInflection` function used anywhere else other than the `InflectionTest` file ? Because I changed the parameters of the function.
- Could you please list all the tests which are required to be added after this optional `<initial>` parsing code has been completed.
- Is there any C file too which I need to update just like this `InflectionTest` ?

BHK4321:
@grhoten 
I did the test these are the results please tell what might be going wrong:


Screenshot 2025-08-29 204358

grhoten:
> Please review this commit and I had some questions while working too:
> 
> * Is the `compareInflection` function used anywhere else other than the `InflectionTest` file ? Because I changed the parameters of the function.

static functions are not available outside of the file. It also compiles. So that implies that the changes are OK.

> * Could you please list all the tests which are required to be added after this optional `<initial>` parsing code has been completed.

As long as there are a couple of tests, that should be sufficient to show the utility of the API.

> * Is there any C file too which I need to update just like this `InflectionTest` ?

I recommend including a test of iinf_createWithDefaults in InflectableStringConceptTest-c.cpp.

> I did the test these are the results please tell what might be going wrong:

`DictionaryLookupFunction::getFeatureValue` likely needs to be updated to query the relevant values from displayValue. It should probably query `DisplayValue::getFeatureValue` or `DisplayValue::getConstraintMap`

inflection/test/resources/inflection/dialog/inflection/es.xml
```diff
@@ -221,4 +221,11 @@
     <test><source definiteness="definite">álgebra añeja</source><result>el álgebra añeja</result></test> <!-- An exception to the gender rule. -->
     <test><source definiteness="definite">añeja álgebra</source><result>la añeja álgebra</result></test>
     <test><source>área</source><result withDemAdj="esta área"/></test>
+
+    <!-- New Tests for stating grammemes ~ bhaskar -->
+    <test><source definiteness="definite">cometa</source><initial gender="masculine"/><result number="singular" gender="masculine">el cometa</result></test>
+    <test><source definiteness="definite">cometa</source><initial gender="feminine"/><result number="singular" gender="feminine">la cometa</result></test>
+    <test><source definiteness="definite">QQQQ</source><initial gender="masculine" number="plural"/><result number="plural" gender="masculine">los QQQQ</result></test>
+    <test><source definiteness="definite">QQQQ</source><initial gender="feminine" number="plural"/><result number="plural" gender="masculine">las QQQQ</result></test>
```
grhoten: These tests look reasonable.

inflection/src/inflection/dialog/DictionaryLookupFunction.cpp
```diff
@@ -126,6 +124,11 @@ SpeakableString* DictionaryLookupFunction::getFeatureValue(const DisplayValue& d
     if (result.empty()) {
         return nullptr;
     }
+    for(const auto& [feature, value]: displayValue.getConstraintMap()) {
+        if (feature.getName() == u"speak") {
+            return new SpeakableString(result, value);
+        }
+    }
     return new SpeakableString(result);
 }
```
BHK4321, line 131: @grhoten 
Was this the expected behavior in this function? If not, could you please point out where I am going wrong?

grhoten: No. Within the DictionaryLookupFunction, you're only returning the grammeme value. This does not need to change.

BHK4321: > DictionaryLookupFunction::getFeatureValue likely needs to be updated to query the relevant values from displayValue. It should probably query DisplayValue::getFeatureValue or DisplayValue::getConstraintMap

Okay sure I wanted to know exactly what I need to do, based on your previous comment I came up with this commit, could you please tell what is expected to make the tests pass so that i could implement the same and further add tests in InfectableStringConceptTest-c.cpp as well to complete this task ?

BHK4321:
Hello!!
@grhoten 
As you can see all tests for the new API for C and C++ are passing as you can see in the Ubuntu build logs , and some error occured in mac os build, I don't know the reason behind it.

inflection/src/inflection/dialog/InflectableStringConcept.cpp
```diff
@@ -52,11 +52,40 @@ SpeakableString* InflectableStringConcept::getFeatureValue(const SemanticFeature
...
+        }
+        if (baseValue) {
+            return baseValue.release();
+        }
+    }
+    const auto& initialConstraintMap = defaultDisplayValue.getConstraintMap();
+    auto initialConstraint = initialConstraintMap.find(feature);
+    if (initialConstraint != initialConstraintMap.end()) {
+        return new SpeakableString(initialConstraint->second);
+    }
```
grhoten: These changes are confusing. Why are these changes needed?

grhoten: More precisely, why is getFeatureValue being called here?  Why is the feature name being checked for gender? What's special about gender or number? Why not grammatical case?

BHK4321:
```cpp
auto displayConstraint = displayConstraintMap.find(feature);
if (displayConstraint != displayConstraintMap.end()) {
    auto numberFeature = npc(getModel())->getFeature(u"number");
    if (feature.getName() == u"gender" && baseValue && numberFeature != nullptr
        && displayConstraintMap.find(*npc(numberFeature)) != displayConstraintMap.end()) {
        return baseValue.release();
    }
    return new SpeakableString(displayConstraint->second);
} 
```

This is the test case which was failing before the changes:
```xml
<test>
  <source definiteness="definite">cometa</source>
  <initial gender="masculine"/>
  <result gender="masculine" number="singular">el cometa</result>
</test>
```

- `InflectableStringConcept::getFeatureValue` asks the default display function for a rendered `DisplayValue`, so feature lookups reflect whatever the display layer injected (articles, inflections, etc.). Spanish determiners add both `gender` and `number` to that map, and the gender can contradict the lexeme default (`la cometa`). 

- To avoid reporting the article’s guess as the noun’s gender, the code detects `feature == "gender"` and, when the number constraint is present too, returns the lexeme’s base value instead of the display-layer value. English articles never set gender, and no language mutates case in this path yet (I am not sure though), so only the gender/number pair needs a guard today.

- `InflectableStringConcept::getFeatureValue` asks the default display function for a rendered `DisplayValue`, so feature lookups reflect whatever the display layer injected (articles, inflections, etc.).

- Spanish determiners add both `gender` and `number` to that map, and the gender can contradict the lexeme default (`la cometa`). 
To avoid reporting the article’s guess as the noun’s gender, the code detects `feature == "gender"` and, when the number constraint is present too, returns the lexeme’s base value instead of the display-layer value, as these conditions when satisfied point to the case when we have an article being inserted.

- Grammatical case isn’t treated specially because the Spanish article logic never synthesizes case information—so there’s no regression to guard against. If a language later injected case-altering determiners we’d need a similar safeguard, but today only the gender/number pair exhibits the mismatch.

BHK4321:
@grhoten 
This final commit resolves the macos build case too by removing the hard coded include path of `xml2` in the `CMakeLists` and using:

```
find_package(LibXml2 REQUIRED)
```
instead to get the include path automatically.

inflection/src/inflection/dialog/InflectableStringConcept.cpp
grhoten: This behavior does not scale, and it does not make sense in my mind. An InflectableStringConcept is just a simplified SemanticConcept. A SemanticConcept should not have this issue. It does not have this kind of logic, and it does not check on specific grammatical category values.

inflection/CMakeLists.txt
```diff
@@ -88,9 +88,7 @@ file(MAKE_DIRECTORY ${INFLECTION_INCLUDE_ROOT})
 
 include(dependICU)
 
-add_library(xml2 INTERFACE IMPORTED GLOBAL)
-set_target_properties(xml2 PROPERTIES IMPORTED_LIBNAME xml2)
-target_include_directories(xml2 INTERFACE ${CMAKE_OSX_SYSROOT}/usr/include/libxml2)
+find_package(LibXml2 REQUIRED)
```
grhoten, line 91: This seems like a good improvement.

inflection/test/src/util/XMLUtils.cpp
```diff
@@ -11,6 +11,7 @@
 #include <inflection/npc.hpp>
 
 std::map<std::u16string, std::u16string> XMLUtils::getAttributes(xmlNodePtr node) {
+    if(node == nullptr) return { };
```
grhoten: Please split this line and use curly braces. It makes it easier to place break points.

BHK4321:
The only test failing is this one:

```xml
<test><source definiteness="definite">QQQQ</source><initial gender="feminine" number="plural"/><result number="plural" gender="masculine">las QQQQ</result></test>
```

This is a spanish-specific issue in which the article has association with gender and the `EsGrammarSynthesizer_CountGenderLookupFunction` looks into the complete displayValue `las QQQQ` and infers the gender grammeme to be feminine based on the articles gender which is overriding the canonical gender of the noun `QQQQ` which is masculine. That is why these two constraints are not matching. So I think this is a locale specific issue so it should be addressed in the synthesizer instead of `DictionaryLookupFunction::getFeatureValue`.

Could you suggest some next-steps?

inflection/src/inflection/dialog/InflectableStringConcept.cpp
BHK4321: Yes, you're right it is just a general constructor and logic, I think the underlying locale specific synthesizer needs to be addressed for the failing test case.

BHK4321:
@grhoten Is the discussion going in the wrong direction or if you could suggest some other way? We only need to address this issue for the failing test

inflection/test/resources/inflection/dialog/inflection/es.xml
grhoten: Let me correct my comment. The structure looks generally good, but the last test is wrong. The initial gender is feminine. So the gender result should be feminine too. This looks like a copy and paste error.

inflection/test/src/inflection/dialog/InflectableStringConceptTest-c.cpp
```diff
@@ -155,6 +155,73 @@ TEST_CASE("InflectableStringConceptTest-c#testExistsAPISpanish")
...
+    REQUIRE(U_SUCCESS(error));
+
+    IDDisplayValue_Constraint defaultConstraints[] = {
+        {u"gender", u"masculine"},
+    };
+
+    auto inflectableConcept = iinf_createWithDefaults(model,
+                                                      sourceString,
+                                                      defaultConstraints,
+                                                      static_cast<int32_t>(sizeof(defaultConstraints) / sizeof(defaultConstraints[0])),
```
grhoten: Please change this calculation to use std::ssize instead. It's a C++20 feature.

inflection/src/inflection/dialog/DictionaryLookupFunction.cpp
grhoten: Can you please revert the change to this file. It looks like it's only changing the empty lines.

inflection/src/inflection/dialog/InflectableStringConcept-c.cpp
```diff
@@ -2,14 +2,18 @@
...
+#include <inflection/npc.hpp>
 #include <inflection/dialog/InflectableStringConcept.hpp>
 #include <inflection/dialog/SemanticFeatureConcept.h>
 #include <inflection/exception/ClassCastException.hpp>
 #include <inflection/util/ULocale.hpp>
+#include <inflection/dialog/SemanticFeature.hpp>
+#include <inflection/dialog/SemanticUtils.hpp>
+#include <inflection/dialog/SemanticFeatureModel.hpp>
 #include <inflection/util/TypeConversionUtils.hpp>
 #include <inflection/util/Validate.hpp>
```
grhoten, line 14: Generally, I've tried to keep the first line the header for the current file, followed by an empty line, sort the headers, and leave npc.h last.

Can you please sort the headers in this section?

inflection/src/inflection/dialog/InflectableStringConcept.cpp
```diff
@@ -63,11 +77,7 @@ ::std::optional<DisplayValue> InflectableStringConcept::getDisplayValue(bool all
 {
     auto defaultDisplayFunction = npc(getModel())->getDefaultDisplayFunction();
     if (defaultDisplayFunction != nullptr && !constraints.empty()) {
-        ::std::map<SemanticFeature, ::std::u16string> constraintMap;
-        if (!value.speakEqualsPrint()) {
-            constraintMap.emplace(*npc(getSpeakFeature()), value.getSpeak());
-        }
-        SemanticFeatureModel_DisplayData displayData({DisplayValue(value.getPrint(), constraintMap)});
+        SemanticFeatureModel_DisplayData displayData({defaultDisplayValue});
```
grhoten: For future reference to self and other reviewers, the DisplayValue constructor with a SpeakableString takes care of the speak usage. It didn't disappear.

grhoten:
> @grhoten Is the discussion going in the wrong direction or if you could suggest some other way? We only need to address this issue for the failing test

Sorry for the delay. The changes are in the right direction now. Once the new comments are addressed, then these changes should be ready to merge.

inflection/src/inflection/dialog/InflectableStringConcept.cpp
BHK4321: @grhoten 
Yes, in the latest commit I did the following change in the InflectableStringConcept::getDisplayValue:
As their are initial constraints as well we might need to pass the constraintMap which was constructed into the defaultDisplayValue. I am not sure, please validate this step, Although the tests are passing.

BHK4321: If this step is correct i will resolve the macos test issue and remove the lines used for debugging so that the PR is ready to merge.

inflection/test/src/inflection/dialog/InflectionTest.cpp
```diff
@@ -2,6 +2,7 @@
  * Copyright 2021-2024 Apple Inc. All rights reserved.
  */
 #include "catch2/catch_test_macros.hpp"
+#include <iostream>
```
grhoten: Please remove the logging changes with iostream. Please use Catch for logging any relevant information. The tests should run clean by default so that it's easier to find relevant information for failing tests.

inflection/src/inflection/dialog/InflectableStringConcept.cpp
```diff
@@ -52,11 +52,15 @@ SpeakableString* InflectableStringConcept::getFeatureValue(const SemanticFeature
...
+    // auto constraintMap = defaultDisplayValue.getConstraintMap();
+    // if(constraintMap.find(feature) != constraintMap.end()) {
+    //     return new SpeakableString(constraintMap[feature]);
+    // }
     auto defaultFeatureFunction = npc(getModel())->getDefaultFeatureFunction(feature);
     if (defaultFeatureFunction != nullptr) {
         const auto displayValueResult = getDisplayValue(true);
         if (displayValueResult) {
-            return npc(defaultFeatureFunction)->getFeatureValue(*displayValueResult, constraints);
+            return npc(defaultFeatureFunction)->getFeatureValue(*displayValueResult, constraints);
```
grhoten: No need for whitespace after the semicolon.

inflection/src/inflection/dialog/InflectableStringConcept.cpp
```diff
@@ -52,11 +52,15 @@ SpeakableString* InflectableStringConcept::getFeatureValue(const SemanticFeature
     if (constraint != nullptr) {
         return new SpeakableString(*npc(constraint));
     }
+    // auto constraintMap = defaultDisplayValue.getConstraintMap();
+    // if(constraintMap.find(feature) != constraintMap.end()) {
+    //     return new SpeakableString(constraintMap[feature]);
+    // }
```
grhoten: Please don't leave commented out code in your changes. It makes it harder to read.

inflection/test/resources/inflection/dialog/inflection/es.xml
```diff
@@ -226,6 +226,6 @@
     <test><source definiteness="definite">cometa</source><initial gender="masculine"/><result number="singular" gender="masculine">el cometa</result></test>
     <test><source definiteness="definite">cometa</source><initial gender="feminine"/><result number="singular" gender="feminine">la cometa</result></test>
     <test><source definiteness="definite">QQQQ</source><initial gender="masculine" number="plural"/><result number="plural" gender="masculine">los QQQQ</result></test>
-    <test><source definiteness="definite">QQQQ</source><initial gender="feminine" number="plural"/><result number="plural" gender="masculine">las QQQQ</result></test>
+    <test><source definiteness="definite">QQQQ</source><initial gender="feminine" number="plural"/><result number="plural" gender="feminine">las QQQQ</result></test>
```
grhoten, line 229: 👍

inflection/test/src/inflection/dialog/InflectionTest.cpp
BHK4321: Yes actually I just wanted to be sure that the changes I did to InflectableStringConcept::getDisplayValue are correct I mean using the constraintMap from defaultDisplayValue so that we use the initial constraints as well, please validate the overall changes to that function from all commits . And then I will do a final cleanup commit.

grhoten: Everything else looks fine, except for the comments that I provided about an hour ago.

BHK4321:
I am not sure about the macos test issue, please look into the same and other than that the code is ready to merge.

grhoten (approved):
The macOS issue is a separate issue. ICU 78 has different behavior, and it requires the tests to be updated. The curly quote has to be switched to an ASCII straight quote.

We will have to update that in a separate pull request.

BHK4321:
@grhoten I think this PR is also ready to be merged. I am really interested in the inflection project and look forward to contribute more in the future.

grhoten (approved):

PR #177, opened 2025-09-10
relax the rules for sr noun inflection

nciric:
Don't require full lexical info about the noun (gender and number) when inflecting with rules. Update tests to reflect that.

Part of issue #172

### DIFF
--- a/inflection/test/resources/inflection/dialog/inflection/sr.xml
+++ b/inflection/test/resources/inflection/dialog/inflection/sr.xml
@@ -22,7 +22,7 @@
     <test><source case="instrumental" pos="noun">Италија</source><result>Италијом</result></test>
     <test><source case="instrumental" pos="noun">авенија</source><result>авенијом</result></test>
     <test><source case="locative" number="plural" gender="feminine" pos="noun">авенија</source><result>авенијама</result></test>
-    <test><source case="vocative" number="singular" gender="masculine" pos="noun">кадија</source><result>кадија</result></test>
+    <test><source case="vocative" pos="noun">кадија</source><result>кадија</result></test>
     <test><source case="vocative" pos="noun">уметница</source><result>уметнице</result></test>
     <test><source case="vocative" pos="noun">птица</source><result>птица</result></test>
     <test><source case="vocative" pos="noun">Стана</source><result>Стано</result></test>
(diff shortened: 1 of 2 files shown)


inflection/test/resources/inflection/dialog/inflection/sr.xml
```diff
@@ -19,24 +19,24 @@
...
-    <test><source case="genitive" number="plural" gender="feminine" pos="noun">гошћа</source><result>гошћа</result></test>
-    <test><source case="genitive" number="plural" gender="feminine" pos="noun">двојка</source><result>двојака</result></test>
-    <test><source case="genitive" number="plural" gender="feminine" pos="noun">битка</source><result>битака</result></test>
+    <test><source case="vocative" pos="noun">кадија</source><result>кадија</result></test>
+    <test><source case="vocative" pos="noun">уметница</source><result>уметнице</result></test>
+    <test><source case="vocative" pos="noun">птица</source><result>птица</result></test>
+    <test><source case="vocative" pos="noun">Стана</source><result>Стано</result></test>
+    <test><source case="vocative" pos="noun">Зора</source><result>Зоро</result></test>
+    <test><source case="vocative" pos="noun">Божа</source><result>Божо</result></test>
+    <test><source case="vocative" pos="noun">Љуба</source><result>Љубо</result></test>
```
grhoten: You can also default to noun.  Every other language assumes noun by default. Every other part of speech is an exception. That's important for quantities.

grhoten (approved):
Looks OK, but you can go further and remove the pos.  Case is the most important part to inflect normally.

inflection/test/resources/inflection/dialog/inflection/sr.xml
nciric: How do you default to POS_NOUN? I looked around other languages but couldn't find an obvious solution in the codebase.

grhoten: Did you try just removing it?

Your dictionary inflector has the following. It means prioritize nouns over adjectives when there is ambiguity.

```
    , dictionaryInflector(::inflection::util::LocaleUtils::SERBIAN(),{
            {GrammemeConstants::POS_NOUN(), GrammemeConstants::POS_ADJECTIVE()},
```

nciric: I did and it just works - so the system will take nouns first, then adjectives as that's the order in the dictionaryInflector. Good to know.

PR #189, opened 2026-05-14
Fix unaligned memory access in memory-mapped file loading

nciric:
This Pull Request addresses crashes (such as SIGILL / Illegal Instruction) caused by unaligned memory access when loading and reading data from memory-mapped files.

## This is a temporary fix, to unblock integration on some platforms. The long term solution is going to adopt structure alignment approach.

The Problem
The current implementation performs direct pointer casting (e.g., _((X_) readCursor)) on data read from memory-mapped files. This assumes that the data in the file is always perfectly aligned to the natural alignment requirements of type X (e.g., 8-byte alignment for int64_t). 

However, binary dictionary files may not always have the necessary padding to guarantee this alignment at every offset. On architectures that strictly enforce data alignment, or when compiler optimizations (such as those utilizing AVX instructions) assume correct alignment, dereferencing these unaligned pointers leads to undefined behavior and runtime crashes.

The Solution
This PR introduces a safer approach that maintains the performance benefits of memory mapping without risking unaligned access crashes:

1.  Safe Primitive Reads: Replaced direct pointer casting with std::memcpy in MemoryMappedFile::readFromCursor. Modern compilers optimize this into efficient single-instruction loads on most platforms, ensuring safety without performance penalty.
2.  UnalignedArray Wrapper: Introduced a lightweight UnalignedArray<X> wrapper to handle array access. It provides an operator[] that safely extracts elements using memcpy, avoiding raw pointer dereference on potentially unaligned addresses in the mapped file.
3.  Refactored Dependents: Updated internal structures like StringArrayContainer and CompressedArray to use this wrapper instead of raw pointers for mapped data.

All unit tests have been verified to pass successfully after these changes.

### DIFF
--- a/inflection/src/inflection/dictionary/DictionaryMetaData_MMappedDictionary.cpp
+++ b/inflection/src/inflection/dictionary/DictionaryMetaData_MMappedDictionary.cpp
@@ -21,7 +21,7 @@ DictionaryMetaData_MMappedDictionary::DictionaryMetaData_MMappedDictionary(const
 
 DictionaryMetaData_MMappedDictionary::DictionaryMetaData_MMappedDictionary(::inflection::util::MemoryMappedFile* memoryMappedRegion, const ::std::u16string& sourcePath)
     : options(npc(memoryMappedRegion)->read<int16_t>())
-    , locale(npc(memoryMappedRegion)->readArray<char>(MAX_LANGUAGE_CODE_LENGTH))
+    , locale(npc(memoryMappedRegion)->readArray<char>(MAX_LANGUAGE_CODE_LENGTH).data_ptr())
     , typesStringContainer(memoryMappedRegion)
     , bitsTypesSingletons(npc(memoryMappedRegion)->read<int8_t>())
     , bitsPropertyMapId(npc(memoryMappedRegion)->read<int8_t>())
@@ -209,7 +209,7 @@ DictionaryMetaData_MMappedDictionary* DictionaryMetaData_MMappedDictionary::crea
     }
 
     // Verify header starts with magic token
-    const char* magicMarker = npc(mappedRegion)->readArray<char>(sizeof(MAGIC_MARKER));
+    const char* magicMarker = npc(mappedRegion)->readArray<char>(sizeof(MAGIC_MARKER)).data_ptr();
     if (strncmp(magicMarker, MAGIC_MARKER, sizeof(MAGIC_MARKER)) != 0) {
         throw ::inflection::exception::IOException(u"Input file " + sourcePath + u" has invalid header");
     }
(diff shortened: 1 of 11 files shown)


nciric:
It seems that problems start when clang is invoked with -O3 option vs O2 or Os.

Crashing with a SIGILL (Illegal Instruction) on x86_64 when using -O3 combined with mmap data loading is a classic indicator of an alignment mismatch intersecting with advanced CPU extensions.While unaligned memory access usually throws a SIGSEGV or SIGBUS on some architectures, it manifests as a SIGILL on x86_64 when dealing with AVX/AVX2/AVX-512 SIMD vectorization.

AVX Alignment Enforcement
When you compile with -O3, Clang aggressively auto-vectorizes loops. If your build system also specifies target architecture flags like -march=native, -march=haswell, or -march=x86-64-v3, Clang will emit advanced Intel/AMD AVX instructions (e.g., vmovaps, vmovdqa).

The Rule: Intel's aligned AVX instructions strictly require the memory address to be aligned to a 32-byte (AVX2) or 64-byte boundary.

The Crash: If an aligned instruction hits an unaligned memory address, the x86 CPU triggers a specific hardware fault that the Linux kernel often translates directly into a SIGILL or SIGSEGV depending on the exact instruction state.

How to Fix It

Fix 1: Align Your File/Data Offsets (Recommended)
If you control the data layout inside the mmap file, pad your headers so that raw numerical arrays always begin at a 64-byte boundary (or at minimum 32-byte boundary).

```cpp
struct FileHeader {
    uint32_t magic;
    uint32_t version;
    char padding[56]; // Pad the header to exactly 64 bytes
};
```

Fix 2: Tell Clang the Data Might Be Unaligned
If you cannot change the file layout, you must explicitly warn Clang's -O3 optimizer that your pointer does not follow strict pointer alignment rules. Use a typedef with the aligned(1) attribute to force the compiler to emit unaligned-safe instructions (vmovups):
```cpp
// Create an unaligned float type
typedef float unaligned_float __attribute__(( some_attributes, aligned(1) ));

// Cast your mmap offset pointer to the unaligned type
char* mmap_ptr = (char*)mmap_data + offset;
unaligned_float* data_array = (unaligned_float*)mmap_ptr;

// Clang will now safely vectorize this loop using unaligned-safe instructions
for (int i = 0; i < size; i++) {
    process(data_array[i]);
}
```

Fix 3: Target a Safe Optimization Local via Pragma
If the loop consuming the mmap data is localized, you can leave your entire project at -O3, but lower the optimization parameters or turn off vectorization exclusively for that single file or function:
```cpp
#pragma clang loop vectorize(disable)
for (int i = 0; i < size; i++) {
    // This loop will execute safely using scalar instructions
    data[i] = ... 
}
```

grhoten:
As per our discussion, I am proposing fix 1., which is what ICU does. The usage of padding can be determined with an option flag in the file. I know that the dictionary has space for such an option bit.

grhoten (approved):
With the understanding that this is a temporary fix, and that a future planned fix will align the data structures with the `-fsanitize=alignment` option confirming the fix.

PR #190, opened 2026-05-14
Fix compilation warnings and missing includes

nciric:
This PR addresses several compilation warnings and missing includes encountered when building with stricter compilers and newer C++ standards (like C++20).

Changes included:
Missing standard headers: Added <cstdlib> for getenv and llabs in ResourceLocator.cpp and NumberConcept.cpp.
Missing ICU headers: Added utf16.h, uchar.h, and utf8.h in files using ICU macros without explicit includes (NumberCleaverIterator.cpp, StringUtils.cpp, StringViewUtils.cpp).
Compiler Warning Fix: Removed an unused variable in ICUTokenExtractorIterator.cpp that triggered a nodiscard warning.
C++20 Compatibility: Added template deduction guides for CompressedArray and MarisaTrie in their respective headers to resolve warnings about class template argument deduction.

These changes help in making the codebase compatible with a wider range of compilers and standards.

All tests passed locally.

### DIFF
--- a/inflection/src/inflection/tokenizer/iterator/ICUTokenExtractorIterator.cpp
+++ b/inflection/src/inflection/tokenizer/iterator/ICUTokenExtractorIterator.cpp
@@ -69,9 +69,8 @@ static void checkinCachedTokenizer(UBreakIterator* tokenizer)
                 gTokenizerCache->push(tokenizer);
                 tokenizer = nullptr;
             }
-            catch (const std::bad_alloc& e) {
+            catch (const std::bad_alloc&) {
                 // Oh well, delete it below
-                e.what();
             }
         }
         if (tokenizer != nullptr) {
--- a/inflection/src/inflection/dialog/NumberConcept.cpp
+++ b/inflection/src/inflection/dialog/NumberConcept.cpp
@@ -11,6 +11,7 @@
 #include <inflection/npc.hpp>
 #include <icu4cxx/NumberFormat.hpp>
 #include <math.h>
+#include <cstdlib>
 
 namespace inflection::dialog {
 
--- a/inflection/src/inflection/dictionary/metadata/CompressedArray.hpp
+++ b/inflection/src/inflection/dictionary/metadata/CompressedArray.hpp
@@ -45,10 +45,6 @@ private:
     CompressedArray& operator=(const CompressedArray& other) = delete;
 };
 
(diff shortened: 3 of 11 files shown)


grhoten:
This code already requires C++20 according to here: https://github.com/unicode-org/inflection/blob/main/inflection/CMakeLists.txt#L35

The code is definitely biased towards clang. I believe that I got the code to run on Linux with g++ once I removed the Core Foundation dependency. Either it's a different version of clang, or something beyond the typical clang++/g++, or another platform beyond the typical Linux.

The header issues are odd, but we can make those changes.

I'm not a fan of the template changes. Can you provide the actual warning? The call site should probably be changed so that the issue doesn't appear.

inflection/src/inflection/tokenizer/iterator/ICUTokenExtractorIterator.cpp
```diff
@@ -69,9 +69,8 @@ static void checkinCachedTokenizer(UBreakIterator* tokenizer)
                 gTokenizerCache->push(tokenizer);
                 tokenizer = nullptr;
             }
-            catch (const std::bad_alloc& e) {
+            catch (const std::bad_alloc&) {
                 // Oh well, delete it below
-                e.what();
             }
```
grhoten, line 74: This was meant to address a different warning about not handling such issues, but whatever. The behavior of ignoring the rare exception is intentional.

nciric:
> This code already requires C++20 according to here: https://github.com/unicode-org/inflection/blob/main/inflection/CMakeLists.txt#L35
> 
> The code is definitely biased towards clang. I believe that I got the code to run on Linux with g++ once I removed the Core Foundation dependency. Either it's a different version of clang, or something beyond the typical clang++/g++, or another platform beyond the typical Linux.
> 
> The header issues are odd, but we can make those changes.
> 
> I'm not a fan of the template changes. Can you provide the actual warning? The call site should probably be changed so that the issue doesn't appear.

The compiler (Clang) emitted -Wctad-maybe-unsupported warnings because Class Template Argument Deduction (CTAD) was used without explicit deduction guides or because the compiler wasn't sure if deduction was intended.

Let's change the call sites.

grhoten (approved):
The changes look good. Do you have more information about the compiler or options used? Perhaps it can be added to the GitHub configurations to test so that this doesn't regress.

PR #191, opened 2026-05-15
fix stable sort

nciric:
_Why stable_sort is better than sort in this case_

The `DictionaryLookupInflector` is responsible for selecting the best inflection candidate for a given word based on a set of constraints. It retrieves potential candidates and sorts them using a comparator that computes a match score.

However, it is common for multiple candidate inflections to have identical match scores. For example, in Hebrew, a single unpointed stem can often correspond to both a past tense verb and an imperative verb with the same gender and number features, yielding an equal match score if those specific features are not part of the disambiguation constraints.

Previously, the code used `std::ranges::sort`, which is an _unstable_ sort. This means that when two candidates have the same score, their relative order in the output is unspecified and non-deterministic. In practice, this led to flaky test failures where the "correct" candidate (as expected by the test suite) would sometimes be selected and sometimes not, depending on the arbitrary order returned by the sort algorithm.

By switching to `std::ranges::stable_sort`, we preserve the relative order of candidate inflections that compare equal. This makes the selection process deterministic and consistent across different runs and environments, effectively resolving the flaky test failures observed in the Hebrew test suite while maintaining the existing prioritization logic.

Another failing example is 'yen' -> 'yens'.

I've checked other reasons for this to happen and this seems like the best answer (I checked different versions of marisa tree, and ICU and nothing mattered).

### DIFF
--- a/inflection/src/inflection/dialog/DictionaryLookupInflector.cpp
+++ b/inflection/src/inflection/dialog/DictionaryLookupInflector.cpp
@@ -89,7 +89,7 @@ inline void traceLogSortedInflectionGrammemes(const ::std::vector<::inflection::
             return DictionaryLookupInflector::compareInflectionGrammemes(inflection1, inflection2, disambiguationGrammemes) < 0;
         };
 
-        std::ranges::sort(inflectionGrammemes, inflectionComparator);
+        std::ranges::stable_sort(inflectionGrammemes, inflectionComparator);
         if (LoggerConfig::isTraceEnabled()) {
             traceLogSortedInflectionGrammemes(inflectionGrammemes, dictionary);
         }


grhoten (changes requested):
This is a reasonable change. Please add at least 1 test highlighting this issue.

grhoten:
Perhaps the test can be added to inflection/test/resources/inflection/dialog/inflection? If it's an existing test failure, can you state which one?

nciric:
> Perhaps the test can be added to inflection/test/resources/inflection/dialog/inflection? If it's an existing test failure, can you state which one?

This is tricky to test in a stable manner. Here's the run with sort vs stable_sort in my environment (and implementation of sort).

1. With std::ranges::sort (Unstable Sort)
The test suite failed with 14 assertions failing across 4 test cases. Here are the specific failures:

"yen" failures (English):
    *   QuantifyTest#testEnglish: Expected 2 yen, but got 2 yens.
    *   QuantifyTest-c#testEnglish: Expected 2 yen, but got 2 yens.
    *   SemanticConceptTest#testEnglishDefaultToSemantic: Expected yen, but got yens.
    *   InflectionTest#testInflections: For locale en source yen with {number=plural}, it expected yen but got yens.

Hebrew failures:
    *   InflectionTest#testInflections: Multiple Hebrew words failed when requesting the feminine gender, typically returning the wrong form (e.g., ending in ה instead of י or vice versa). Examples include:
        *   Source מחק: Expected מחקי, got מחקה
        *   Source בחר: Expected בחרי, got בחרה
        *   Source התכונן: Expected התכונני, got התכוננה
        *   Source לחץ: Expected לחצי, got לחצה
        *   Source התחבר: Expected התחברי, got התחברה
        *   ... and several others.

2. With std::ranges::stable_sort (Stable Sort)
The test suite passed successfully with zero failures (cached and verified without cache).

Existing test failure: Yes, it causes failures in QuantifyTest, SemanticConceptTest, and InflectionTest.
Specific cases:
    *   English: Failures on the word "yen" (pluralizing to "yens" instead of staying "yen").
    *   Hebrew: Failures on multiple words like מחק (eraser/delete), בחר (choose), התכונן (prepare), etc., when inflecting for feminine gender.

grhoten (approved):
If the existing tests were failing before this change, then that's good enough for me. Though it would be nice if a pipeline could be set up with this failing configuration so that we don't regress.

nciric:
> If the existing tests were failing before this change, then that's good enough for me. Though it would be nice if a pipeline could be set up with this failing configuration so that we don't regress.

My guess is the implementation difference in sort. I would need to figure out which std lib we use to set up the environment.

PR #192, opened 2026-05-19
Avoid class template argument deduction (CTAD) to fix warnings

nciric:
_Description_
This PR avoids using Class Template Argument Deduction (CTAD) for `inflection::util::Finally` and `inflection::dictionary::metadata::MarisaTrie` by explicitly specifying template arguments or using `decltype` for lambdas.

_Motivation_
In some environments (e.g. with strict warning flags like `-Wctad-maybe-unsupported` enabled), using CTAD without explicit deduction guides can trigger compiler warnings or errors. 

Instead of adding deduction guides to the headers, this PR changes the local invocations to avoid deduction entirely, making the code more robust across different compiler configurations and avoiding the need for deduction guides.


grhoten (approved):
Definitely falls into the pedantic category.

nciric:
> Definitely falls into the pedantic category.

Let's just say that all warnings are converted to errors in some places, pedantic or not :).

PR #193, opened 2026-06-17
Implement German Compound Word Tokenizer and Fix Capitalization during Inflection

nciric:
This PR implements the German compound word tokenizer (`DeTokenizer`) and integrates it into the inflection engine. It also resolves key issues related to Fugenelemente matching and case sensitivity during inflection lookup for split compounds.

### 1. German Tokenizer Configuration (`DeTokenizer`)
*   Implemented `DeTokenizer` and registered it in [TokenizerFactory.cpp](file:///usr/local/google/home/cira/code/inflection/inflection/src/inflection/tokenizer/TokenizerFactory.cpp).
*   Configured German to use the new tokenizer and the compiled dictionary (`/de/tokenizer.tokd`) in [config_de.properties](file:///usr/local/google/home/cira/code/inflection/inflection/resources/org/unicode/inflection/tokenizer/config_de.properties).
*   Configured German Fugenelemente (`s`, `es`, `e`, `n`, `en`, `er`, `ens`) in the new `DeDictionaryTokenizerConfig`.

### 2. German Capitalization Fix in Inflection Lookup
*   **The Problem:** When a compound noun is split by the tokenizer, the head word becomes lowercase (e.g., `Ehefrau` -> `Ehe` + `frau`). During inflection lookup in the case-sensitive inflection dictionary, this lowercase `frau` matched the lowercase pronoun `frau` instead of the capitalized noun `Frau`, leading to incorrect POS tagging and failed inflection.
*   **The Fix:** Updated [DeGrammarSynthesizer_DeDisplayFunction.cpp](file:///usr/local/google/home/cira/code/inflection/inflection/src/inflection/grammar/synthesis/DeGrammarSynthesizer_DeDisplayFunction.cpp) to check if a split head word was part of a capitalized compound. If so, it capitalizes the head word (e.g., `frau` -> `Frau`) before looking it up in the inflection dictionary, and restores it to lowercase in the final reassembled string.

### 3. Fugenelemente Matching Bug Fix
*   Fixed a bug in [DictionaryTokenizerConfig.cpp](file:///usr/local/google/home/cira/code/inflection/inflection/src/inflection/tokenizer/dictionary/DictionaryTokenizerConfig.cpp) where reversed keys were being checked against non-reversed sets in `isPositiveFugenelement` and `isReplaceableFugenelement`, which broke multi-character Fugenelemente matching.

### 4. Realistic German Tokenizer Dictionary with Flags
*   Generated a realistic 197k-word `tokenizer.dictionary` from `dictionary_de.lst` and Wiktionary German frequency list.
*   Set appropriate flags in the dictionary to improve split quality:
    *   `isNoCompound` for proper nouns and abbreviations (prevents them from splitting).
    *   `isNoAtomic` for pronouns and articles (prevents them from being used as segments).
    *   `isSegment` for short content words (allows words like `Ei` to be recognized as segments).
*   Boosted frequencies for `television` (prevents false splits) and `haupt` (forces `Hauptstraße` to split).

## Verification
*   Added unit tests in [TokenizerTest.cpp](file:///usr/local/google/home/cira/code/inflection/inflection/test/src/inflection/tokenizer/TokenizerTest.cpp) covering German compound splitting (`Hauptstraße`, `Waldweg`, `Schlossgasse`, `Marktplatz`, `Kaiserring`, `Sonnenallee`, `Arbeitszimmer`, `Kindergarten`).
*   Ran `make check` - all 260 test cases passed (262636 assertions).

### DIFF
--- a/inflection/src/inflection/grammar/synthesis/DeGrammarSynthesizer_DeDisplayFunction.cpp
+++ b/inflection/src/inflection/grammar/synthesis/DeGrammarSynthesizer_DeDisplayFunction.cpp
@@ -472,6 +472,33 @@ std::optional<::std::pair<::std::u16string, ::std::u16string>> DeGrammarSynthesi
         iteratedToken = npc(iteratedToken)->getPrevious();
     }
 
+    // German nouns are capitalized. When a compound noun is split, the head word
+    // becomes lowercase (e.g., 'Ehefrau' -> 'Ehe' + 'frau'). We must capitalize it
+    // for dictionary lookup to avoid missing the noun entry or matching a lowercase
+    // homonym with a different POS (e.g., matching pronoun 'frau' instead of noun 'Frau').
+    ::std::u16string headWordStr(headWord);
+    bool headWordCapitalized = false;
+    if (!headWordStr.empty() && u_islower(headWordStr[0])) {
+        // Find the start of the compound
+        const inflection::tokenizer::Token* compoundStartToken = headWordToken;
+        auto iterated = headWordToken;
+        while (iterated != nullptr) {
+            if (iterated->isWhitespace()) {
+                break;
+            }
+            compoundStartToken = iterated;
+            iterated = npc(iterated)->getPrevious();
+        }
+        if (compoundStartToken != nullptr && !compoundStartToken->getValue().empty() && u_isupper(compoundStartToken->getValue()[0])) {
+            ::std::u16string capitalized = inflection::util::StringViewUtils::capitalizeFirst(headWordStr, ::inflection::util::LocaleUtils::GERMAN());
+            int64_t capGrammemes = 0;
+            if (dictionary.getCombinedBinaryType(&capGrammemes, capitalized) != nullptr && (capGrammemes & dictionaryNoun) != 0) {
+                headWordStr = capitalized;
+                headWordCapitalized = true;
+            }
+        }
+    }
+
     std::optional<::std::u16string> inflectedHeadWord;
     std::optional<::std::u16string> inflectedDependentWord;
 
@@ -479,7 +506,7 @@ std::optional<::std::pair<::std::u16string, ::std::u16string>> DeGrammarSynthesi
     if (dependentToken != nullptr && npc(dependentToken)->isSignificant()) {
         auto const& dependentWord = dependentToken->getValue();
 
-        const auto inflectionResult = inflect2Words(dependentWord, headWord, constraints, enableInflectionGuess);
+        const auto inflectionResult = inflect2Words(dependentWord, headWordStr, constraints, enableInflectionGuess);
 
         if (!inflectionResult) {
             // inflection has failed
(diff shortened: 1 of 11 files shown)


inflection/test/src/inflection/tokenizer/TokenizerTest.cpp
```diff
@@ -409,6 +409,19 @@ TEST_CASE("TokenizerTest#testDutchDictionary")
     compareValueTokens(tokenizer->createTokenChain(u"bluetoothinstelling"), {u"", u"bluetooth", u"instelling", u""});
 }
 
+TEST_CASE("TokenizerTest#testGermanDictionary")
```
grhoten, line 412: Please move this test to `inflection/test/resources/inflection/tokenizer/de.xml`. It's easier to maintain there.

inflection/src/inflection/grammar/synthesis/DeGrammarSynthesizer_DeDisplayFunction.cpp
```diff
@@ -472,14 +472,41 @@ ::std::optional<::std::u16string> DeGrammarSynthesizer_DeDisplayFunction::inflec
...
+        }
+        if (compoundStartToken != nullptr && !compoundStartToken->getValue().empty() && u_isupper(compoundStartToken->getValue()[0])) {
+            ::std::u16string capitalized = inflection::util::StringViewUtils::capitalizeFirst(headWordStr, ::inflection::util::LocaleUtils::GERMAN());
+            int64_t capGrammemes = 0;
+            if (dictionary.getCombinedBinaryType(&capGrammemes, capitalized) != nullptr && (capGrammemes & dictionaryNoun) != 0) {
+                headWordStr = capitalized;
+                headWordCapitalized = true;
+            }
+        }
+    }
```
grhoten, line 500: This implementation detail should not be in this class. The tokenizer is supposed to handle it. Also your tokenizer dictionary seems to be correctly normalized to lowercase.  Also the lexical dictionary is case insensitive for lookup.  Can you explain this logic? If you're using AI, I think it's making some assumptions.

inflection/src/inflection/tokenizer/dictionary/DictionaryTokenizerConfig.hpp
```diff
@@ -32,6 +32,9 @@ class inflection::tokenizer::dictionary::DictionaryTokenizerConfig
     ::std::vector<::std::u16string_view> replacements {  };
     ::std::vector<::std::u16string_view> fugenelements_negative {  };
     ::std::set<::std::u16string, std::less<>> fugenelements {  };
+    ::std::set<::std::u16string, std::less<>> fugenelements_positive_reversed {  };
+    ::std::set<::std::u16string, std::less<>> fugenelements_replaceable_reversed {  };
```
grhoten, line 36: Warning: The data alignment issue allowed me to break the binary compatibility and the assumption about using everything reversed. I'm switching it non-reversed in the upcoming changes.  In any case, I don't think that these should have been separate. These changes should be using fugenelements_positive and fugenelements_replaceable.

inflection/src/inflection/tokenizer/locale/de/DeTokenizer.hpp
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright 2016-2024 Apple Inc. All rights reserved.
```
grhoten, line 2: Are you sure about these copyrights?

grhoten (changes requested):
The tokenizer.dictionary seems reasonable, but I didn't investigate it in depth. Though you should reconsider the inclusion of shorter words, like "a". Perhaps consider excluding such words to make the dictionary smaller.

The `config_de.properties` changes seem correct. The additions to `inflection/src/inflection/tokenizer/locale/de/` and `inflection/src/inflection/tokenizer/TokenizerFactory.cpp` seem reasonable.

The rest of the changes are unexpected. No changes should have been needed for `DeGrammarSynthesizer_DeDisplayFunction.cpp` because it already works with a tokenizer that decompounds words. I wouldn't expect changes to `DictionaryTokenizerConfig` either. So I'd like an explanation of those changes in detail.

I know that nouns and adjectives differ by letter casing in German, but I suspect that this detail doesn't align with how this code works.  This code assumes that the letter casing may be wrong, or in sentence case (e.g. an adjective at the beginning of a sentence). So it doesn't strictly adhere to letter casing norms of German. It does a lot of case insensitive matching. If you do provide a proper letter cased, it will try to use that style as a default, but it's not strictly required.

PR #194, opened 2026-06-27
Inflection-188 Support PronounConcept in icu::message2 Inflection-187 Support SemanticConcept in icu::message2 for custom vocabulary Inflection-186 Support the print and speak lines from SpeakableString in icu::message2 Inflection-185 Support lists in icu::message2 Inflection-183 Support quantify in icu::message2 Inflection-178 Provide quantity support for Serbian Inflection-117 Support inflection functionality into icu::message2 Inflection-87 Integrate Unicode Inflection into Unicode Message Format Support more languages

### LINKED ISSUE #87: Integrate Unicode Inflection Into Message Format 2
This is a GSoC Project idea.

Difficulty/Size: Higher

The goal is to create a way for the [MessageFormatter in ICU4C](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1message2_1_1MessageFormatter.html) to leverage the [Unicode Inflection](https://github.com/unicode-org/inflection) project for creating grammatically correct sentences.  These are currently 2 independent projects, and Unicode Inflection is only available with C and C++ API at this time.  Unicode Inflection supports all of the use cases in the expected outcomes, but MessageFormatter has no way to use such functionality at this time.

Below is background material for the Unicode Inflection concepts with 3 different syntaxes that are not the MessageFormat 2 syntax. Unicode Inflection supports all of the use cases in the expected outcomes.

- [UTW 2023 Automatic Grammar Agreement in Message Formatting](https://www.youtube.com/watch?v=C2e7hYIkqoM)
- [S12T1 Authoring Grammatically Correct Conversational Templates for Siri](https://www.youtube.com/watch?v=emlIWUTaJFM)
- [Let's Come To An Agreement About Our Words :: IMUG 2017.02.16](https://www.youtube.com/watch?v=KclVxxHX26k)

Here is some additional background information between these projects.

- [UTW 2024 Solving Inflection](https://www.youtube.com/watch?v=Z9M8SwIbvMM)
- [UTW 2024 MessageFormat 2 Technical Preview: Where Are We Now?](https://www.youtube.com/watch?v=4jucYXE42_s)

Expected Outcomes

1. A minimum viable integration can generate these messages depending on the grammatical properties of the object being inserted into the sentence.
   1. English
       1. The {object} is on
       1. The {object} are on
   1. French
       1. La {object} est allumée
       1. Les {object} sont allumées
       1. L'{object} est allumé
       1. Le {object} est allumé
       1. Les {object} sont allumés
1. Bonus integrations include the following topics
   1. Support quantities. E.g. 1 foot/2 feet
   1. Support lists. E.g. An object, and a table
   1. Support pronouns, especially in Spanish and Arabic. E.g. Here is {pronoun} location.
   1. Support spoken text. E.g. one foot/two feet.
   1. Support inflecting words. E.g. mouse + plural → mice
   1. Support SemanticConcept to allow custom inflections.
   1. Support inflection alternative when the word being inserted into the sentence is not in the lexical dictionary.  E.g. 😀 isn’t in the lexical dictionary. So you may want to use “L(a|e) 😀est allumé(e)” in French.

**Skills**

- Required: C/C++
- Required: Ability to use cmake for building the code
- Preferred: Love of languages.

### LINKED ISSUE #113: Derive sound properties from Wikidata for Korean and Turkish
Right now, the sound properties are usable for vowel-start, vowel-end, consonant-start and consonant-end. These values are contained in inflection/tools/dictionary-parser/src/main/resources/org/unicode/wikidata/P898.properties.

New regular expressions are needed for the following properties for Korean and Turkish.

rieul-end
front-unround
front-round
back-unround
back-round

See the following for reference.
https://en.wikipedia.org/wiki/Rieul_(hangul)
https://en.wikipedia.org/wiki/Close_front_unrounded_vowel
https://en.wikipedia.org/wiki/Close_front_rounded_vowel
https://en.wikipedia.org/wiki/Close_back_unrounded_vowel
https://en.wikipedia.org/wiki/Close_back_rounded_vowel

The rieul sound is for latin letters only because the Hangul script is phonetic enough to derive algorithmically.

### LINKED ISSUE #117: Support inflection functionality into icu::message2
icu::message2 provide two intreface to integrate additional functionality into MF2
icu::message2::FormatFactory and
icu::message2::SelectorFactory

This issue aim to add concrete implementations of that two interface from inflectino library so the caller can easily adopt the functionality inside inflection into ICU4C MF2.

This a sub task of #61

### LINKED ISSUE #178: Provide quantity support for Serbian
Problem:
Demo program at https://github.com/nciric/test-inflection/blob/main/quantity.cc outputs the following for two words (one in lexicon and the other one not).

Run it with a word from a dictionary, like девојка, the output is
print: 1 девојка
speak: једна девојка
print: 2 девојке
speak: две девојке
print: 1 девојци
speak: једна девојци (NOTE: wrong case for number, RBNF update needed)

Run it with a word outside dictionary (rule based), like боја, the output is
print: 1 боја
speak: 1 боја
print: 2 боја
speak: 2 боја
print: 1 боји
speak: 1 боји

It's worse in two aspects:
1. Speak doesn't "kick in" as I always get 1 or 2 as digits
2. It seems not to inflect (except for the last 2, which is even more confusing)

Two things have to happen to fix this issue:
1. ICU needs updated RBNF forms
2. Common concept factory needs to be updated, e.g. here's Russian example inflection/src/inflection/dialog/language/RuCommonConceptFactory.cpp

### LINKED ISSUE #183: Support quantify in icu::message2
The MessageFormat integration should include integration with `CommonConceptFactory::quantify`.  Here are some possible examples of what that may look like.

en_US (US English)
Example | Number | Printed Result | Spoken Result
|----|----|----|----|
`{$number :quantify unit=\|new message\|}`|101|101 new messages|one hundred one new messages
`{$number :quantify unit=\|megaparsec\|}`|101|101 megaparsecs|one hundred one megaparsecs
`{$number :quantify unit=$unit}`|101|101 kilometers|one hundred one dogs

de (German)
Example | Number | Printed Result | Spoken Result
|----|----|----|----|
`{$number :quantify unit=\|Film\| case=dative}`|1|1 Film|einem Film
`{$number :quantify unit=\|Film\| case=dative}`|2|2 Filmen|zwei Filmen

The German case would be invalid for English, and it should be marked as invalid in some way. I'm not sure how to support the `formattedNumber` argument.

### LINKED ISSUE #185: Support lists in icu::message2
The `SemanticConceptList` provides a way to format an "and/conjunction" list, an "or/disjunction" list, and a generic list. The generic list could be used for an adjective list. The ListTest.cpp contains a lot of examples of how it can be used.

Given an array of [SemanticFeatureConceptBase](https://unicode-org.github.io/inflection/classinflection_1_1dialog_1_1SemanticFeatureConceptBase.html), you should be able to apply various text transformations or inflection operations. For example, I could make each item in a list definite or indefinite. Some items may use the default behavior of InflectableStringConcept, but some custom inflections may want to use SemanticConcept.

I could query a list for the gender value and use it in a selector operation.

A `SemanticConceptList` is special because it can use context sensitive words. The "and" or "or" may vary depending on the adjacent words.

The following optional parameters need to be supported.

- type or list: This could be "and" or "conjuction", or "or" or "disjunction". If it's nothing, the default is used.
- beforeFirst: You might want to add a demonstrative pronoun or possessive pronoun here, like a `PronounConcept`.
- itemPrefix: 
- itemSuffix: You might want to add a possessive pronoun, like a `PronounConcept`.
- afterFirst: 
- itemDelimiter: Most people that this would be a ", ", but for adjective lists, it could be just a space in some languages.
- beforeLast: 
- afterLast: When used in some East Asian languages, like Chinese, this may be used for adjective lists, where you may add . When the list is not empty, this will get formatted.

Ideally, the list should allow chained transformations on each item, like title casing, and quoting, or making each genitive, or making each use a possessive pronoun. The possessive pronouns are important for languages like Arabic, Malay, Indonesian, and Finnish.

Adding a demonstrative pronoun would require adding the equivalent of "this" or "that" to each item, and the choice of the demonstrative pronoun would be sensitive to the gender or grammatical number of team item. So a list like [cat, dogs] with "this" being applied to each item would have "this cat, and these dogs".

### LINKED ISSUE #186: Support the print and speak lines from SpeakableString in icu::message2
This may be architecturally hard to support in icu::message2, but it would be helpful to support formatting of SpeakableString. There are cases where you need to print the text, or speak the text, or both. It's impossible to perfectly get ambiguous text perfectly spoken, or spoken text perfectly transcribed. You sometimes need to give a little guidance or provide an explicit way to speak or print text.

[SSML](https://www.w3.org/TR/speech-synthesis11/) provides a way to annotate text for proper speaking of the printed text, but it works best when the text is fully known at build time. It works less well for variables. Think of SSML as something similar to CSS for text. You're providing the style of how to speak instead of style to print.

As an example, how would someone pronounce "$1.23". If I'm referring to a US speaker in the US region talking about US dollar quantities, I'd say "one dollar and twenty three cents". If the same person were actually talking about Canadian dollars, I'd more likely want to disambiguate the pronunciation as "one Canadian dollar and twenty three cents". The opposite would happen if I swapped roles and currencies.

The normal process for the Unicode Inflection usage is to format and modify each segment as each part is being generated. It's not to format it twice, or print the whole thing and then speak the whole. It's to do both at the same time, and concatenate the SpeakableString.

### LINKED ISSUE #187: Support SemanticConcept in icu::message2 for custom vocabulary
`SemanticConcept` support should be added to icu::message2. This provides a way to perform custom inflections when the default inflection behavior is ambiguous, a language is unsupported by Unicode Inflection or precise inflection behavior is needed for custom terms are needed.

Support will need to hook into `SemanticFeatureModel`. There likely needs to be some sort of file format to handle this custom vocabulary that is fed into the SemanticFeatureModel.

Examples can be seen in `SemanticConceptTest.cpp`.

### LINKED ISSUE #188: Support PronounConcept in icu::message2
There should be a way to support `PronounConcept`. Since [Message Format 2](https://messageformat.unicode.org/docs/quick-start/) does not provide a way to structure data with properties, I guess it will have to be a separate variable. Supporting it completely should take the following into consideration.

1. Be able to specify a hard coded pronoun, like "their", and allow it to be inflected to "his" or "her" or something like that depending on the gender of a person meant for the pronoun.
2. If a PronounConcept has a default gender associated with it, be able to inflect the pronoun like the other objects derived from [SemanticFeatureConceptBase](https://unicode-org.github.io/inflection/classinflection_1_1dialog_1_1SemanticFeatureConceptBase.html). So you could inflect "I" to "we" if inflected to be plural. This is less natural for a translator to use because it requires linguistic knowledge. A more natural way to use it would be to use a known static pronoun, and inflect it based on `PronounConcept` being a vessel for storing such information.
3. Be able to use a `referencedConcept` with `toSpeakableString` to handle the pronouns that depend on the grammatical properties of the `referencedConcept`. So instead of using the gender of the possessor, you use the gender of the possessed. If you look at the pronoun csv files, this data is represented as `dependency=`.
4. Be able to use a `PronounConcept` in a selector. You should be able to query the gender of the pronoun as necessary.

PronounConcept is in reference to people. It is not meant to be used for demonstrative pronouns, like "this" or "that". It's not meant for interrogative pronouns, like "who" or "what". It is meant to be used for human in the first person (the speaker/creator of the message), the second person (the message receiver/audience), the third person (other people), and other objects (animals, or inanimate objects).

grhoten:
Fixes #188
Fixes #187
Fixes #186
Fixes #185
Fixes #183
Fixes #178
Fixes #117
Fixes #113
Fixes #87

The changes address the following issues.

* The existing lexical dictionaries are updated.
* Language support
  * 23 new languages are being supported.
  * 3 languages perform longest suffix guessing when a word is missing from a lexical dictionary.
  * Several tests have been commented out due to missing lexeme data from Wikidata.
  * Several languages have adopted the new RBNF changes for improved grammatical agreement of quantities. The HEAD changes from CLDR are required for these changes, which requires the upcoming ICU 79 release.
  * A lot of data is being added.
* Various performance improvements. Some changes are aligned with ICU performance improvements.
* The structures are data aligned. It follows the ICU design principle for data alignment. This is a binary breaking change, which is OK for a pre-1.0 release.
* There is a blind port to Windows. It’s untested, but it minimizes the effort to support testing on Windows in GitHub.
* Message Format 2 support significantly updated.
  * The old data driven FormatTest has been converted into Message2Test
  * Support added for InflectableStringConcept, SemanticConcept, SemanticConceptList, PronounConcept, NumberConcept, and quantify have been added.
  * Support for SSML for conversion from SpeakableString. This is the best way to shoehorn in the functionality.
  * This coverage is mostly complete. This support should probably be reviewed with the MF2 team.
  * This is an API breaking change, which is OK for a pre-1.0 release.
* The dictionary-parser now has passing tests.
* The Serbian and Malayalam grammar synthesizers are roughly the same, but they could adopt some new internal frameworks.
* And of course various bug fixes.

The version of ICU needs to be updated before the pull request checks can pass. The [Language List](https://github.com/unicode-org/inflection/wiki/Language-list) should be updated once this pull request is merged.


inflection/src/inflection/dialog/language/SerboCroatianCommonConceptFactory.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <inflection/dialog/SpeakableString.hpp>
+#include <inflection/util/Validate.hpp>
+#include <inflection/npc.hpp>
+#include <cmath>
+
+namespace inflection::dialog::language {
+
+// Croatian (hr) and Serbian (sr) share Bosnian-Croatian-Serbian numeral-noun agreement: the 2-4
+// paucal takes the genitive singular and 5+ takes the genitive plural.
+SerboCroatianCommonConceptFactory::SerboCroatianCommonConceptFactory(const ::inflection::util::ULocale& language)
```
grhoten, line 17: @nciric You may be interested in this.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
grhoten, line 1: The data driven tests in this directory has a lot of examples of how to integrate Unicode Inflection into MF2.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+ Copyright 2026 Unicode Incorporated and others. All rights reserved.
+-->
+<!DOCTYPE message2test SYSTEM "message2test.xsd">
+<message2test locale="en-US">
+    <test>
+        <params>
+            <param name="object" type="String"><value>lights</value></param>
+        </params>
+        <source>The {$object} {is :inflect number=$object} on.</source>
+        <print>The lights are on.</print>
```
grhoten, line 12: This is how you inflect a word based on another word.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+    </test>
+    <test>
+        <params>
+            <param name="object" type="String"><value>lights</value></param>
+        </params>
+        <source>{$object :inflect definiteness=definite} {is :inflect number=$object} on.</source>
+        <print>the lights are on.</print>
+    </test>
```
grhoten, line 20: Add "the" in front of a word. This is boring in English. It's more interesting in the rest of the European languages.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+    <test>
+        <params>
+            <param name="object" type="String"><value>light</value></param>
+        </params>
+        <source>.local $num = {$object :inflect to=number} .match $num
+plural {{The {$object} are on}}
+* {{The {$object} is on}}
+</source>
+        <print>The light is on</print>
+    </test>
```
grhoten, line 30: This is the verbose way to implement this functionality. I'm not a fan of it, but it shows that the property can be queried for a more complex phrase.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+        </params>
+        <source>I didn't find {$relationship :inflect definiteness=indefinite} in your contacts. What is your {$relationship :inflect case=genitive} name?</source>
+        <print>I didn't find a brother in your contacts. What is your brother’s name?</print>
+    </test>
+    <test>
+        <params>
+            <param name="relationship" type="String"><value>aunt</value></param>
+        </params>
+        <source>I didn't find {$relationship :inflect definiteness=indefinite} in your contacts. What is your {$relationship :inflect case=genitive} name?</source>
+        <print>I didn't find an aunt in your contacts. What is your aunt’s name?</print>
```
grhoten, line 73: Now you can use the correct indefinite article in English!

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+    </test>
+    <test>
+        <params>
+            <param name="oridinalNum" type="Number"><value>4</value></param>
+        </params>
+        <source>Your {$oridinalNum :numeral withStyle=asWords withVariant=ordinal} meeting has been canceled</source>
+        <print>Your fourth meeting has been canceled</print>
```
grhoten, line 80: Fun with RBNF ordinals.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+    </test>
+    <test>
+        <params>
+            <param name="count" type="Number"><value>4</value></param>
+        </params>
+        <source>You have {$count :numeral withStyle=asSpokenWords} messages</source>
+        <print>You have 4 messages</print>
+        <speak>You have four messages</speak>
```
grhoten, line 88: These print/speak values are emulated with SSML. See https://www.w3.org/TR/speech-synthesis11/ for details.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+        <source>Your meeting is in {$unit :quantify withValue=$n} from now</source>
+        <print>Your meeting is in 1 day from now</print>
+    </test>
+    <test>
+        <params>
+            <param name="n" type="Number"><value>2</value></param>
+            <param name="unit" type="String"><value>day</value></param>
+        </params>
+        <source>Your meeting is in {$unit :quantify withValue=$n} from now</source>
+        <print>Your meeting is in 2 days from now</print>
```
grhoten, line 104: This is a simple example of how quantities work.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+    </test>
+    <test>
+        <params><param name="p" type="String"><value>she</value></param></params>
+        <source>.local $g = {$p :pronoun to=gender} .match $g
+feminine {{feminine phrase}}
+masculine {{masculine phrase}}
+* {{other phrase}}
+</source>
+        <print>feminine phrase</print>
+    </test>
```
grhoten, line 170: Some simplified ways to handle pronouns.

inflection/test/resources/inflection/message2/en_US.xml
```diff
@@ -0,0 +1,195 @@
...
+        <params>
+            <param name="np" type="PronounConcept">
+                <value constraint="third singular nominative">xe</value>
+                <value constraint="third singular accusative">xem</value>
+                <value constraint="third singular genitive">xyr</value>
+            </param>
+        </params>
+        <source>{$np :pronoun case=accusative} {$np :pronoun}</source>
+        <print>xem xe</print>
+    </test>
```
grhoten, line 194: Custom pronoun.

inflection/test/resources/inflection/message2/es_MX.xml
```diff
@@ -0,0 +1,119 @@
...
+        <print>gato y gata</print>
+    </test>
+    <test>
+        <params>
+            <param name="p" type="String"><value>mío</value></param>
+            <param name="obj" type="String"><value>casas</value></param>
+        </params>
+        <source>{$p :pronoun withReferent=$obj}</source>
+        <print>mías</print>
+    </test>
```
grhoten, line 118: Now we're getting fancier with the possessive pronouns because this is the gender of the object and not of the person.

inflection/test/resources/inflection/message2/ru_RU.xml
```diff
@@ -1,33 +1,53 @@
...
+<message2test locale="ru-RU">
     <test>
         <params>
-            <param name="number" type="Number"><value>1</value></param>
+            <param name="n" type="Number"><value>1</value></param>
             <param name="unit" type="String"><value>километр</value></param>
         </params>
-        <source><quantity value="number" unit="unit"/> <quantity value="number" unit="unit" inflect="genitive"/> <quantity value="number" unit="unit" inflect="accusative"/> <quantity value="number" unit="unit" inflect="dative"/> <quantity value="number" unit="unit" inflect="instrumental"/> <quantity value="number" unit="unit" inflect="prepositional"/></source>
+        <source>{$unit :quantify withValue=$n} {$unit :quantify withValue=$n case=genitive} {$unit :quantify withValue=$n case=accusative} {$unit :quantify withValue=$n case=dative} {$unit :quantify withValue=$n case=instrumental} {$unit :quantify withValue=$n case=prepositional}</source>
         <print>1 километр 1 километра 1 километр 1 километру 1 километром 1 километре</print>
```
grhoten, line 13: A "simple" Russian quantity.

inflection/test/resources/inflection/message2/ru_RU.xml
```diff
@@ -1,33 +1,53 @@
...
+                <value constraint="neuter nominative plural">nominative,plural</value>
+                <value constraint="neuter instrumental plural">instrumental,plural</value>
+                <value constraint="neuter accusative plural">accusative,plural</value>
+                <value constraint="neuter dative plural">dative,plural</value>
+                <value constraint="neuter genitive plural">genitive,plural</value>
+                <value constraint="neuter prepositional plural">prepositional,plural</value>
+            </param>
+        </params>
+        <source>{$unit :inflect gender=neuter case=genitive number=singular}</source>
+        <print>genitive,singular</print>
```
grhoten, line 49: Simple SemanticConcept for custom vocabulary. You have to register this stuff with a SemanticFeatureModel.

inflection/test/resources/inflection/message2/ru_RU.xml
```diff
@@ -46,7 +66,7 @@
                 <value constraint="neuter prepositional plural">prepositional,plural</value>
             </param>
         </params>
-        <source><quantity value="number" unit="unit"/> <quantity value="number" unit="unit" inflect="genitive"/> <quantity value="number" unit="unit" inflect="accusative"/> <quantity value="number" unit="unit" inflect="dative"/> <quantity value="number" unit="unit" inflect="instrumental"/> <quantity value="number" unit="unit" inflect="prepositional"/></source>
+        <source>{$unit :quantify withValue=$number} {$unit :quantify withValue=$number case=genitive} {$unit :quantify withValue=$number case=accusative} {$unit :quantify withValue=$number case=dative} {$unit :quantify withValue=$number case=instrumental} {$unit :quantify withValue=$number case=prepositional}</source>
         <print>1 nominative,singular 1 genitive,singular 1 accusative,singular 1 dative,singular 1 instrumental,singular 1 prepositional,singular</print>
         <speak>одно nominative,singular одного genitive,singular одно accusative,singular одному dative,singular одним instrumental,singular одном prepositional,singular</speak>
```
grhoten, line 71: Now we're getting fancy with a SemanticConcept. Notice how the number pronunciation changes.

inflection/test/resources/inflection/message2/ru_RU.xml
```diff
@@ -90,8 +110,31 @@
                 <value constraint="neuter prepositional plural">prepositional,plural</value>
             </param>
         </params>
-        <source><quantity value="number" unit="unit"/> <quantity value="number" unit="unit" inflect="genitive"/> <quantity value="number" unit="unit" inflect="accusative"/> <quantity value="number" unit="unit" inflect="dative"/> <quantity value="number" unit="unit" inflect="instrumental"/> <quantity value="number" unit="unit" inflect="prepositional"/></source>
+        <source>{$unit :quantify withValue=$number} {$unit :quantify withValue=$number case=genitive} {$unit :quantify withValue=$number case=accusative} {$unit :quantify withValue=$number case=dative} {$unit :quantify withValue=$number case=instrumental} {$unit :quantify withValue=$number case=prepositional}</source>
         <print>5 genitive,plural 5 genitive,plural 5 genitive,plural 5 dative,plural 5 instrumental,plural 5 prepositional,plural</print>
         <speak>пять genitive,plural пяти genitive,plural пять genitive,plural пяти dative,plural пятью instrumental,plural пяти prepositional,plural</speak>
```
grhoten, line 115: This is the kind of stuff that should replace the CLDR measurement units. There are numerous units handled by CLDR. Unfortunately, CLDR doesn't integrate well with unknown measurement units. CLDR measurement units doesn't use RBNF to correctly pronounce a number.

Yes, many Slavic languages are this complicated.

inflection/tools/dictionary-parser/src/test/java/org/unicode/wikidata/ParseWikidataTest.java
```diff
@@ -74,8 +77,8 @@ public void caseInsensitiveGrammemeTest() {
 
     @Test
     public void lexiconParserTest() throws Exception {
-        String[] args = {"--inflection-types", "noun,adjective,proper-noun", "--inflections", inflectionalFile,
-                "--dictionary", dictionaryFile, lexiconSource};
+        String[] args = {"--inflection-types", "noun,adjective,proper-noun", "--add-sound", "vowel-start",
+                "--inflections", inflectionalFile, "--dictionary", dictionaryFile, lexiconSource};
```
grhoten, line 81: Yay! The tool tests pass now!

inflection/tools/genExemplars/genExemplars.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+static void checkArgument(bool failureCondition, string_view message) {
+    if (failureCondition) {
+        cerr << message << endl;
+        cerr << USAGE_STRING << endl;
+        exit(-1);
+    }
+}
+
+int main(int argc, const char* const argv[])
```
grhoten, line 196: This is a new tool to generate exemplar words. It generates a file that matches the longest suffix of an unknown word, and maps it to an equivalent word with the same suffix. The popular suffix wins. Only the popular suffixes are generated, and it can take a little bit of time to generate this file.

inflection/test/src/inflection/grammar/ExemplarTest.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    std::u16string pattern;
+    std::u16string result;
+    inflection::util::StringViewUtils::convert(&pattern, std::string_view(patternPtr, patternSize));
+    inflection::util::StringViewUtils::replace(&result, pattern, std::u16string_view(u"\0", 1), u"");
+    return result;
+}
+
+// TODO enable this test once the suffix files have been regenerated with the latest data.
+TEST_CASE("ExemplarTest#testExemplarsInDictionary", "[.]")
```
grhoten, line 73: This exemplar test has to be reenabled at a future time. The suffix files with the exemplar words need to be regenerated with the words that are actually in the Wikidata lexemes.

inflection/resources/org/unicode/inflection/locale/supported-locales.properties
```diff
@@ -3,27 +3,50 @@
...
 locale.group.sv=sv_SE
+locale.group.ta=ta_IN
+locale.group.te=te_IN
 locale.group.th=th_TH
 locale.group.tr=tr_TR
+locale.group.uk=uk_UA
+locale.group.ur=ur_PK
 locale.group.vi=vi_VN
 locale.group.yue=yue_CN
 locale.group.zh=zh_CN,zh_TW,zh_HK
```
grhoten, line 52: This file advertises all of the new languages.

inflection/tools/scripts/check.sh
```diff
@@ -0,0 +1,23 @@
...
+# Examples:
+#   tools/scripts/check.sh
+#   tools/scripts/check.sh build
+#   tools/scripts/check.sh build-claude "-l nl InflectionTest#testInflections"
+
+BUILD_DIR="${1:-build}"
+shift 2>/dev/null
+TEST_OPTS="$*"
+
+JOBS="$(getconf _NPROCESSORS_ONLN)"
```
grhoten, line 17: This script isn't strictly needed. It's only helpful to trust this test operation with a parallel build while using agentic coding tools. It avoids the prompt.

inflection/src/inflection/dictionary/DictionaryMetaData_MMappedDictionary.hpp
```diff
@@ -70,7 +70,7 @@ class INFLECTION_INTERNAL_API inflection::dictionary::DictionaryMetaData_MMapped
     int8_t bitsPropertyValueMapValuesSize {  };
     int32_t wordsToTypesSingletonsSize {  };
     ::inflection::dictionary::metadata::MarisaTrie<uint64_t> wordsToDataTrie {::std::map<std::u16string_view, uint64_t>()};
-    inflection::util::MemoryMappedFile::UnalignedArray<int64_t> wordsToTypesSingletons;
+    const int64_t* const wordsToTypesSingletons {  };
```
grhoten, line 73: This is now aligned. This temporary workaround has been removed, and the binary file format version has been bumped accordingly. The other file formats were also modified to align its data structures. This isn't a problem on ARM generally, but it can be an issue when certain compiler options are used.

I used the cmake `ALIGNMENT_TEST=ON` option when testing this originally with cmake.

inflection/tools/buildStringMap/buildStringMap.cpp
```diff
@@ -0,0 +1,293 @@
...
+    }
+    writeHeader(out, 1, fromSet, toSet);
+    combinedTrie.write(out);
+}
+
+static constexpr int32_t MINIMUM_SIZE_FOR_DUAL_TRIE = 256;
+static constexpr char USAGE_STRING[] =
+    "Usage: buildStringMap INFILE OUTFILE";
+
+int main(int argc, const char * const argv[]) {
```
grhoten, line 208: This tool generates a bidirectional table for strings based on 1-2 tries. Think of this like how the transliterator works or how ICU converters work (but with strings instead of bytes).

It's currently used for mapping suffixes to exemplar words, but it's a little more general purpose. The file format allows roundtripping, fallback, and reverse fallback. It ensures that mappings are not cyclic beyond round trip mappings (the equals sign), and are not one way.

These mappings are valid:
```
a<->b
a->b<->c
a<->b<-c
a->b<->c<-d
```
These mappings are not valid:
```
a->b
a<-b
a->b->c
```

inflection/test/src/inflection/dialog/DialogThreadSafetyTest.cpp
```diff
@@ -14,72 +14,95 @@
...
         minimumNumberOfWords *= 100;
     }
 
+    const inflection::dialog::SemanticFeatureModel* currModel = nullptr;
+    std::atomic finished(false);
+    std::barrier barrier(processorCount + 1);
+    for (int32_t count = 0; count < processorCount; count++) {
+        results[count].reserve(minimumNumberOfWords);
+        threads.emplace_back(testInflection, &currModel, &phrases, &constraints, &semanticFeatures, &barrier, &finished, count, &results[count]);
+    }
```
grhoten, line 90: We're reusing the threads and modernizing the style of thread safety checking.

inflection/src/inflection/util/ArrayUtils.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include <inflection/util/fwd.hpp>
+#include <compare>
+#include <iterator>
+#include <string_view>
+
+/**
+ * @brief Provides utility functions for use arrays.
+ * @details Provides utility functions for use arrays.
+ */
+class inflection::util::ArrayUtils {
```
grhoten, line 15: This helps reduce the heap usage a fair amount. This allows you to perform a binary search in an arbitrary sorted constexpr C data array, and it's verified as sorted at **compile** time. You get a compiler error if it's unsorted.

inflection/src/inflection/grammar/synthesis/GrammemeConstants.hpp
```diff
@@ -4,104 +4,112 @@
 #pragma once
 
 #include <inflection/grammar/synthesis/fwd.hpp>
-#include <string>
 
 class inflection::grammar::synthesis::GrammemeConstants final
```
grhoten, line 8: This is now a header only class. It avoids permanently caching these strings on the heap, and there is no thread safe initialization to be concerned about.

inflection/src/inflection/message2/MF2Factory.hpp
```diff
@@ -56,20 +55,50 @@
...
-   * to register a custom selector factory.
-   */
-  static icu::message2::SelectorFactory* CreateSelectorFactory();
+      /**
+       * Create an implementation of icu::message2::FormatterFactory*, based on the
+       * infleciton library, which can be passed to
+       * icu::messsage2::MFFunctionRegistry::Builder::adoptFormatter
+       * to register a custom formatter factory.
+       */
+      static icu::message2::FormatterFactory* createInflectFormatterFactory(OutputMode mode = OutputMode::PLAIN);
```
grhoten, line 74: By default, we format as plain text. If you want to retain the spoken form, then it can be outputted as SSML. This is an API breaking change.

As a reminder, this is a pre-1.0 release. So API stability guarantees are not there yet.

inflection/cmake/runTests.cmake
```diff
@@ -1,20 +1,31 @@
...
-set(RUN_COMMAND "${LIBRARY_PATH_NAME}=\${${LIBRARY_PATH_NAME}}:${DYLD_LIBRARY_PATH} $TEST_RUN_COMMAND ${EXECUTABLE} $TEST_OPTS")
-execute_process(
-        COMMAND sh -c "echo Test command: ${RUN_COMMAND}"
-)
-execute_process(
-        COMMAND sh -c "${RUN_COMMAND}"
-        OUTPUT_FILE /dev/stdout
-        ERROR_FILE /dev/stderr
+
+if(WIN32)
```
grhoten, line 9: If someone could run this on Windows, that would be helpful. The default is to print in color, but on Windows, we don't run the `sh` shell. I don't know if this test runs in color. This can be verified by manually running this from the command line outside of the makefiles.

grhoten (commented):
Here are some highlights of these changes.

inflection/src/inflection/message2/MF2Factory.cpp
```diff
@@ -168,39 +326,622 @@ void InflectionSelector::selectKey(
...
+
+    std::unique_ptr<inflection::dialog::NumberConcept> num;
+    if (toFormat.getType() == UFMT_DOUBLE) {
+        num = std::make_unique<inflection::dialog::NumberConcept>(
+            toFormat.getDouble(errorCode), locale, locale);
+    }
+    else {
+        num = std::make_unique<inflection::dialog::NumberConcept>(
+            toFormat.getInt64(errorCode), locale, locale);
+    }
```
grhoten, line 449: We recently had a meeting in ICU around this topic. There are times that formatting a number and speaking a number need to be a bit different in such scenarios. This should not be doubling up the locale usage like this. When formatting context allows this value to be decoupled, we should modify these lines.

This is a temporary workaround for now.

inflection/test/resources/inflection/dialog/inflection/fi.xml
```diff
@@ -0,0 +1,3220 @@
...
+    <test><source number="plural" case="partitive">suomi</source><result>suomia</result></test>
+    <test><source number="singular" case="nominative">Lahti</source><result>Lahti</result></test>
+    <!-- FIXME: <test><source number="singular" case="genitive">Lahti</source><result>Lahden</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="partitive">Lahti</source><result>Lahtea</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="inessive">Lahti</source><result>Lahdessa</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="elative">Lahti</source><result>Lahdesta</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="illative">Lahti</source><result>Lahteen</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="adessive">Lahti</source><result>Lahdella</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="ablative">Lahti</source><result>Lahdelta</result></test> -->
+    <!-- FIXME: <test><source number="singular" case="allative">Lahti</source><result>Lahdelle</result></test> -->
```
grhoten, line 212: Due to the lack of Wikidata lexeme coverage, there are numerous tests that are failing, especially for the new languages. These tests have a FIXME tag on each failing test. There are over a thousand of these data coverage issues, and the vast majority of them are in Finnish.

For all languages, adding/fixing the missing data is usually the best way to fix the issue.

For Finnish, Polish, and Russian, adding the missing words mentioned in the suffix tables will allow numerous words to be supported en masse. Addressing that data coverage issue would allow the Exemplar test to be enabled for that language.

Occasionally, the issue is not a data coverage issue, and it's an algorithm or data structure issue. There are a handful of these issues.  The Norwegian and Dutch failures may fall into this category, but I'm not sure yet.

nciric (commented):
qq about "The existing lexical dictionaries are updated."

I assume no lexemes were added to the dictionaries that don't already exist in Wikidata. If that's not the case, we'll have a problem next time we sync with Wikidata as it will override items.

inflection/resources/org/unicode/inflection/dictionary/dictionary_ro.lst
```diff
@@ -0,0 +1,243 @@
+Atena: proper-noun inflection=15
+Belgia: singular direct definite feminine proper-noun inflection=10
+Belgiei: singular oblique definite feminine proper-noun inflection=10
+Belgii: singular oblique indefinite feminine proper-noun inflection=10
+Bruxelles: singular direct indefinite neuter proper-noun inflection=14
+Franța: singular direct indefinite feminine proper-noun inflection=8
+Japonia: singular direct definite feminine proper-noun inflection=1
+Kатар: singular direct indefinite neuter proper-noun inflection=4
```
nciric: This cyrillic -  probably a bug in Wikidata.

inflection/resources/org/unicode/inflection/dictionary/dictionary_ro.lst
```diff
@@ -0,0 +1,243 @@
...
+venit: past participle verb
+veniti: plural indicative second present verb
+vii: singular indicative second present verb
+vin: singular plural indicative first third present verb
+vine: singular indicative third present verb
+vorbesc: singular indicative first present verb
+îmi: singular dative pronoun
+șpur: singular direct indefinite masculine noun inflection=c
+șpuri: plural direct indefinite masculine noun inflection=c
+Виетнам: singular direct indefinite neuter proper-noun inflection=4
```
nciric: All words from here to the end are Cyrillic - I don't think we want that for Romanian.

inflection/src/inflection/analysis/RussianExposableMorphology.cpp
```diff
@@ -1,113 +1,57 @@
...
-    // If Word all uppercase -> Make inflection all upper case
-    if (inflection::util::StringViewUtils::isAllUpperCase(word)) {
-        ::std::u16string upperCaseInflectedWord;
-        inflection::util::StringViewUtils::uppercase(&upperCaseInflectedWord, inflectedWord, ::inflection::util::LocaleUtils::RUSSIAN());
-        inflectedWord = upperCaseInflectedWord;
-    }
-    return {inflectedWord};
+    static constexpr const char16_t* const doNotInflect[] = {
+        // This must be in sorted order.
+        u"августа",
```
nciric, line 39: Why are month names hardcoded?

grhoten:
> qq about "The existing lexical dictionaries are updated."
> 
> I assume no lexemes were added to the dictionaries that don't already exist in Wikidata. If that's not the case, we'll have a problem next time we sync with Wikidata as it will override items.

I don't quite understand the question. I did not manually edit the data derived from Wikidata, if that is what you mean.

The data is derived like this, and this is how you update the lexical dictionaries.
```
cd inflection/tools/dictionary-parser
./ParseWikidata --all ../../../wikidata-20260708-lexemes.json
```

Note: I have to update this data once more before the final review.

inflection/src/inflection/dictionary/Inflector_InflectionPattern.cpp
```diff
@@ -51,14 +45,26 @@ Inflector_Inflection inflection::dictionary::Inflector_InflectionPattern::getInf
...
         *this,
         int32_t(suffixIdx),
-        inflectorDictionary.grammemePatterns[grammemesIdx]
+        *(inflectorDictionary.grammemePatterns + grammemesIdx)
     };
 }
 
 bool Inflector_InflectionPattern::containsSuffix(std::u16string_view suffix) const
 {
+    const auto suffixID = inflectorDictionary.inflectionSuffixes.getIdentifierIfAvailable(suffix);
```
nciric, line 54: This is where we "guess" word inflection if not in dictionary by suffix in inflectional_xx.xml?

inflection/src/inflection/dialog/language/SerboCroatianCommonConceptFactory.cpp
nciric: Is there something I need to change or is Serbian going to pick it up by default (I can look at Croatian for guidance).

nciric (approved):

nciric:
> > qq about "The existing lexical dictionaries are updated."
> > I assume no lexemes were added to the dictionaries that don't already exist in Wikidata. If that's not the case, we'll have a problem next time we sync with Wikidata as it will override items.
> 
> I don't quite understand the question. I did not manually edit the data derived from Wikidata, if that is what you mean.
> 
> The data is derived like this, and this is how you update the lexical dictionaries.
> 
> ```
> cd inflection/tools/dictionary-parser
> ./ParseWikidata --all ../../../wikidata-20260708-lexemes.json
> ```
> 
> Note: I have to update this data once more before the final review.

I understand the process, my questions was about data source:
1. Is it all from wikidata
2. Or some of it was directly added to the dictionary without being added to wikidata

grhoten:
> my questions was about data source:
> 
> 1. Is it all from wikidata
> 2. Or some of it was directly added to the dictionary without being added to wikidata

It's from Wikidata with some filtering and augmentation from ParseWikidata (dictionary-parser). That tool does augment some data that isn't in Wikidata, like adding sound properties or some other grammemes. The filter properties there does reorder, remove, and merge some lexemes.

It's all automated. No manual edits were performed.

The sound properties could be included into Wikidata by adding phonetic information to many words. That's hard to derive when you only need a subset of the data. So right now, that data is primarily separate. If Wikidata has phonetic information on a form, it will be parsed and the tool will derive the important information.

inflection/src/inflection/dictionary/Inflector_InflectionPattern.cpp
grhoten: No, check Polish, Finnish, and Russian. They use a suffix file of exemplar words. It's mostly automatically generated. Some manual edits are performed on the derived suffix files. Unfortunately, the suffix files are from a more complete lexical dictionary. They should probably be rederived from Wikidata, but that should wait till after the data in Wikidata has better coverage. The `ExemplarTest#testExemplarsInDictionary` test has a TODO to work on this task.

Right now, it has partial word coverage, which is why the `ExemplarTest#testExemplarsInDictionary` test fails.

inflection/src/inflection/analysis/RussianExposableMorphology.cpp
grhoten: That's a good question. I think this may be a misnomer from a previous version of this code.

It should probably be renamed as isNotGuessable or something like that because it's only used in guessInflection. Though if the words are complete, then it's just dead code/data. This needs to be revisited before merging.

grhoten:
> Is there something I need to change or is Serbian going to pick it up by default (I can look at Croatian for guidance).

I recommend reviewing `QuantifyTest#testSerboCroatian` for accuracy. If there are some grammatical cases besides the nominative form that you want to add, those grammatical case tests can be added. You can look at Russian for some examples.

The `SerboCroatianCommonConceptFactory` class is registered in LocalizedCommonConceptFactoryProvider for Serbian. It's not automatically registered.

inflection/src/inflection/analysis/RussianExposableMorphology.cpp
nciric: The overall change looks good - once this is fixed (and maybe Romanian dictionary) feel free to land.

inflection/resources/org/unicode/inflection/dictionary/dictionary_ro.lst
nciric: I checked with @mihnita and he said the use of Cyrillic is not  expected.

grhoten: Well, it's not completely wrong. There is some history and politics around this topic that is reflected in the Wikidata, which I won't elaborate here. I've added an option to filter out `ro-md` for Romanian, and the data is now filtered out.

inflection/src/inflection/analysis/RussianExposableMorphology.cpp
grhoten: After further review, I think this is fine as is. It looks like if you try to change these genitive forms to accusative for dates, they stay unchanged according to one of the tests. Since this part is refactor, and it was already here, I recommend leaving this as-is. I'm fine with revisiting this separately with a Russian expert.

nciric (approved):

inflection/src/inflection/dialog/language/SerboCroatianCommonConceptFactory.cpp
grhoten: I don't think that there is anything left to change in this class. Croatian is using the same class now. Both are registered in LocalizedCommonConceptFactoryProvider.cpp.

PR #197, opened 2026-07-20
Add secondary longest-suffix exemplar guessing for Serbian

nciric:
I added secondary longest-suffix exemplar guessing for Serbian nouns when rule-based guessing doesn't inflect a word missing from the dictionary.

Key changes:
- Created `suffix_sr.csv` mapping common Serbian Cyrillic noun endings (`-ак`, `-њак`, `-ац`, `-ник`, `-ач`, `-ар`, `-бус`, `-ан`, `-ош`, `-иша`, `-ица`, `-ија`, `-ка`, `-о`, `-бло`, `-е`, `-ње`, `-ме`) to Cyrillic dictionary exemplars.
- Updated `SrGrammarSynthesizer_SrDisplayFunction` to fall back to `suffixToExemplar` + `inflectExemplar` if `inflectWithRule` returns the uninflected word.
- Added 20 Cyrillic test cases in `sr.xml` covering unknown words with diverse endings.


inflection/resources/org/unicode/inflection/exemplar/suffix_sr.csv
```diff
@@ -0,0 +1,19 @@
+ак,=,изузетак
```
grhoten, line 1: Was this file manually generated or generated from genExemplars? It seems a little short.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -159,8 +164,37 @@ ::inflection::dialog::DisplayValue *SrGrammarSynthesizer_SrDisplayFunction::getD
...
         displayString = inflectFromDictionary(constraints, displayString);
     } else if (enableInflectionGuess) {
-        // Let's use rule based inflection for nouns. Assume lemma is singular, nominative.
-        displayString = inflectWithRule(constraints, displayString);
+        // Primary guess: Rule based inflection for nouns.
+        ::std::u16string ruleResult = inflectWithRule(constraints, displayString);
+        if (ruleResult != displayString) {
+            displayString = ruleResult;
+        } else {
+            // Secondary guess: Longest suffix exemplar guessing.
```
grhoten, line 172: This is an interesting combination. Why not go all in for one or the other? What is the advantage of one over the other?

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
```diff
@@ -159,8 +164,37 @@ ::inflection::dialog::DisplayValue *SrGrammarSynthesizer_SrDisplayFunction::getD
...
+            auto countString = GrammarSynthesizerUtil::getFeatureValue(constraints, numberFeature);
+            auto caseString = GrammarSynthesizerUtil::getFeatureValue(constraints, caseFeature);
+            auto genderString = GrammarSynthesizerUtil::getFeatureValue(constraints, genderFeature);
+            std::vector<::std::u16string> constraintsVec;
+            if (!countString.empty()) constraintsVec.push_back(countString);
+            if (!caseString.empty() && caseString != GrammemeConstants::CASE_NOMINATIVE) constraintsVec.push_back(caseString);
+            if (!genderString.empty()) constraintsVec.push_back(genderString);
```
grhoten, line 179: I would have expected these values to already be derived. Why are they being queried here?

Also please don't inline if statements. It makes it hard to debug, and get decent code coverage metrics.

grhoten (commented):
If you're going this route, you might want to enable `ExemplarTest#testExemplarsInDictionary` to ensure that your new table is valid. The TODO can be to enable it for the other languages.

The other suffix tables are a work in progress. Either the missing words need to be added, or the table needs to be regenerated.

inflection/resources/org/unicode/inflection/exemplar/suffix_sr.csv
grhoten: FYI the other tables are mostly generated with some small manual tweaking.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: Rule based inflection is going to give the correct answer for words not in the dictionary (with minor exceptions).
Some of the groups are hard to implement, or require more grammatical information to do properly - the longest suffix match is replacement for those for now. I expect the results to be occasionally incorrect, so it's the least preferred approach of the 3 (dictionary, rules, suffix guess) wrt correctness, but easy to implement :).

inflection/resources/org/unicode/inflection/exemplar/suffix_sr.csv
nciric: This was manually generated - I'll look into genExemplars, thanks for the pointer.

inflection/src/inflection/grammar/synthesis/SrGrammarSynthesizer_SrDisplayFunction.cpp
nciric: So I actually went and implemented the other 3 cases and it wasn't that bad.
I hit a snag on this one as the suffix matching would need to be hand curated to avoid some of the special cases and with the PR #198 this PR is not needed.

Let's discuss it tomorrow.

PR #207, opened 2026-08-12
Inflection-206 Fix missing ICU UTF-16 header includes, C++17/20 CTAD guides, and empty INFLECTION_ROOT path resolution

### LINKED ISSUE #206: Fix missing ICU UTF-16 header includes, C++17/20 CTAD guides, and empty INFLECTION_ROOT path resolution
### Overview
This issue addresses compiler compatibility and runtime path resolution improvements across `inflection`:

1. **Explicit `<unicode/utf16.h>` includes**:
   Several C++ source files use ICU UTF-16 code point macros (`U16_PREV`, `U16_GET`, `U16_LENGTH`, `U16_NEXT`) without explicitly including `<unicode/utf16.h>`. On strict compiler toolchains that do not transitively include ICU sub-headers, compilation fails.

2. **C++17/20 Class Template Argument Deduction (CTAD) Guides**:
   Template classes (`Finally`, `CompressedArray`, `MarisaTrie`) and test code (`DialogThreadSafetyTest.cpp`) rely on template argument deduction. On toolchains compiling with strict C++17/20 warning options (such as `-Wctad-maybe-unsupported`), compiler warnings or errors are triggered. Adding explicit deduction guides for these classes and specifying explicit template arguments (`std::atomic<bool>`, `std::barrier<>`) ensures clean compilation.

3. **`ResourceLocator.cpp` Relative Path Resolution**:
   In `ResourceLocator::getRootForLocale()`, when the `INFLECTION_ROOT` environment variable is unset/empty, string concatenation `INFLECTION_ROOT_DIRECTORY() + u"/" + U16(INFLECTION_DATA_ROOT_DIR) + u"/inflection"` evaluates with an unintended leading slash (`u"/..."`). This converts relative `INFLECTION_DATA_ROOT_DIR` paths into absolute filesystem root paths. Updating `getRootForLocale` to check if `INFLECTION_ROOT_DIRECTORY()` is empty preserves intended relative directory pathing.

---

### Proposed Changes

- **Header Includes**: Add `#include <unicode/utf16.h>` to:
  - `src/inflection/tokenizer/ControlCleaver.cpp`
  - `src/inflection/grammar/synthesis/HeGrammarSynthesizer_WithConditionalHyphen.cpp`
  - `src/inflection/grammar/synthesis/KoGrammarSynthesizer_ParticleResolver.cpp`
  - `src/inflection/grammar/synthesis/TrGrammarSynthesizer_TrDisplayFunction.cpp`
  - `src/inflection/lang/StringFilterUtil.cpp`
  - `src/inflection/tokenizer/TokenUtil.cpp`
  - `src/inflection/util/StringViewUtils.cpp`
  - `src/inflection/util/UnicodeSetUtils.cpp`

- **CTAD Guides & Test Updates**:
  - `src/inflection/util/Finally.hpp`: Add deduction guide `template <typename Function> Finally(Function) -> Finally<Function>;`
  - `src/inflection/dictionary/metadata/CompressedArray.hpp`: Add deduction guide `template <typename T> CompressedArray(const std::vector<T>&) -> CompressedArray<T>;`
  - `src/inflection/dictionary/metadata/MarisaTrie.hpp`: Add deduction guide `template <typename T> MarisaTrie(const std::map<std::u16string_view, T>&) -> MarisaTrie<T>;`
  - `test/src/inflection/dialog/DialogThreadSafetyTest.cpp`: Specify explicit template parameters for `std::atomic<bool>` and `std::barrier<>`.

- **Path Resolution**:
  - `src/inflection/util/ResourceLocator.cpp`: Check if `INFLECTION_ROOT_DIRECTORY()` is empty before prepending `u"/"`.

nciric:
Fixes #206

### Summary of Changes

This PR improves compiler portability and C++ template deduction compatibility across `inflection`:

1. **Explicit ICU `<unicode/utf16.h>` Includes**:
   - Explicitly `#include <unicode/utf16.h>` in 8 source files (`ControlCleaver.cpp`, `TokenUtil.cpp`, `StringViewUtils.cpp`, `UnicodeSetUtils.cpp`, `StringFilterUtil.cpp`, and Hebrew/Korean/Turkish grammar synthesizers) that use UTF-16 code point macros (`U16_PREV`, `U16_GET`, `U16_LENGTH`, `U16_NEXT`).
   - Resolves build failures on hermetic toolchains and IWYU-strict environments where ICU sub-headers are not transitively exported by `<unicode/uchar.h>` or `<unicode/ustring.h>`.

2. **C++17/20 Template Argument Deduction (CTAD) Guides & Test Fixes**:
   - Added user-defined deduction guides for `Finally`, `CompressedArray`, and `MarisaTrie` to allow clean class template deduction from constructor arguments (e.g. lambdas, `std::vector<T>`, `std::map<..., T>`).
   - Explicitly specified `std::atomic<bool>` and `std::barrier<>` in `DialogThreadSafetyTest.cpp`.
   - Resolves compilation failures and warnings when building under `-std=c++20` with LLVM `libc++` or when `-Wctad-maybe-unsupported` is enabled.

### Verification
- Builds and unit tests pass cleanly across Linux, macOS, and Windows CI.

### DIFF
--- a/src/inflection/grammar/synthesis/HeGrammarSynthesizer_WithConditionalHyphen.cpp
+++ b/src/inflection/grammar/synthesis/HeGrammarSynthesizer_WithConditionalHyphen.cpp
@@ -8,6 +8,7 @@
 #include <inflection/util/StringViewUtils.hpp>
 #include <unicode/uchar.h>
 #include <unicode/uscript.h>
+#include <unicode/utf16.h>
 
 namespace inflection::grammar::synthesis {
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ if(MSVC)
 else()
     add_compile_options(-Wall -Weffc++ -Wextra)
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-        add_compile_options(-Wglobal-constructors -Wexit-time-destructors -Wweak-vtables -Wvla-extension)
+        add_compile_options(-Wglobal-constructors -Wexit-time-destructors -Wweak-vtables -Wvla-extension -Wctad-maybe-unsupported)
     endif()
 
     # Improve code security
--- a/src/inflection/grammar/synthesis/KoGrammarSynthesizer_ParticleResolver.cpp
+++ b/src/inflection/grammar/synthesis/KoGrammarSynthesizer_ParticleResolver.cpp
@@ -13,6 +13,7 @@
 #include <inflection/tokenizer/Token_Tail.hpp>
 #include <inflection/tokenizer/Tokenizer.hpp>
 #include <inflection/tokenizer/TokenizerFactory.hpp>
+#include <unicode/utf16.h>
 #include <inflection/npc.hpp>
 #include <memory>
 #include <unicode/uchar.h>
--- a/src/inflection/grammar/synthesis/TrGrammarSynthesizer_TrDisplayFunction.cpp
+++ b/src/inflection/grammar/synthesis/TrGrammarSynthesizer_TrDisplayFunction.cpp
@@ -13,6 +13,7 @@
 #include <inflection/dictionary/PhraseProperties.hpp>
 #include <inflection/grammar/synthesis/GrammemeConstants.hpp>
 #include <inflection/grammar/synthesis/TrGrammarSynthesizer.hpp>
+#include <unicode/utf16.h>
 #include <inflection/util/LocaleConstants.hpp>
 #include <inflection/util/LocaleUtils.hpp>
 #include <inflection/util/StringViewUtils.hpp>
--- a/src/inflection/lang/StringFilterUtil.cpp
+++ b/src/inflection/lang/StringFilterUtil.cpp
@@ -12,6 +12,7 @@
 #include <icu4cxx/UnicodeSet.hpp>
 #include <unicode/ulocdata.h>
 #include <unicode/uscript.h>
+#include <unicode/utf16.h>
 #include <map>
 #include <mutex>
 
--- a/src/inflection/tokenizer/ControlCleaver.cpp
+++ b/src/inflection/tokenizer/ControlCleaver.cpp
@@ -11,6 +11,7 @@
 #include <inflection/tokenizer/TokenUtil.hpp>
 #include <inflection/npc.hpp>
 #include <unicode/uchar.h>
+#include <unicode/utf16.h>
 
 namespace inflection::tokenizer {
 
--- a/src/inflection/tokenizer/TokenUtil.cpp
+++ b/src/inflection/tokenizer/TokenUtil.cpp
@@ -7,6 +7,7 @@
 #include <inflection/util/UnicodeSetUtils.hpp>
 #include <inflection/npc.hpp>
 #include <unicode/uchar.h>
+#include <unicode/utf16.h>
 
 namespace inflection::tokenizer {
 
--- a/src/inflection/util/StringViewUtils.cpp
+++ b/src/inflection/util/StringViewUtils.cpp
@@ -11,6 +11,7 @@
 #include <unicode/uchar.h>
 #include <unicode/ustring.h>
 #include <unicode/utf8.h>
+#include <unicode/utf16.h>
 
 namespace inflection::util {
 
--- a/src/inflection/util/UnicodeSetUtils.cpp
+++ b/src/inflection/util/UnicodeSetUtils.cpp
@@ -5,6 +5,7 @@
 #include <inflection/util/StringViewUtils.hpp>
 #include <inflection/util/StringUtils.hpp>
 #include <inflection/npc.hpp>
+#include <unicode/utf16.h>
 
 namespace inflection::util {
 
--- a/test/src/inflection/dialog/DialogThreadSafetyTest.cpp
+++ b/test/src/inflection/dialog/DialogThreadSafetyTest.cpp
@@ -82,8 +82,8 @@ TEST_CASE("DialogThreadSafetyTest#testInflect", "[multithreaded]")
     }
 
     const inflection::dialog::SemanticFeatureModel* currModel = nullptr;
-    std::atomic finished(false);
-    std::barrier barrier(processorCount + 1);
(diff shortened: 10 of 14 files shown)


src/inflection/util/ResourceLocator.cpp
```diff
@@ -90,7 +90,11 @@ ::std::u16string inflection::util::ResourceLocator::getRootForLocale(const infle
     if (!dataRegistrationServicePath.empty()) {
         return StringViewUtils::to_u16string(dataRegistrationServicePath);
     }
-    return INFLECTION_ROOT_DIRECTORY() + u"/" + U16(INFLECTION_DATA_ROOT_DIR) + u"/inflection";
+    const auto& root = INFLECTION_ROOT_DIRECTORY();
+    if (root.empty()) {
+        return ::std::u16string(U16(INFLECTION_DATA_ROOT_DIR)) + u"/inflection";
+    }
+    return root + u"/" + U16(INFLECTION_DATA_ROOT_DIR) + u"/inflection";
```
grhoten: The original line was purposefully there by default. When `INFLECTION_ROOT` is undefined, a relative path from the library/program is derived and not the current working directory. You do not want the current working directory. That could lead to some security concerns, and will likely undo the effort of ResourceLocator::getDirectoryFromEnvironment. Using the OS is usually better than the current working directory, which is too arbitrary to use. The absolute path is normally better when the directory of the library could not be deduced.

If ResourceLocator::getDirectoryFromEnvironment were ever empty, it would likely be because some non-standard configuration was used (though the correct directory structure for Windows still needs to be figured out).  At that point, the user moved the files to an unexpected relative path to the library, and we would need help.

The current behavior is trying to avoid the mess from how `ICU_DATA` needed to be defined for many years. Most people got it wrong. Most people don't move the relative path of executables around normally though.

Please don't make this change.

src/inflection/grammar/synthesis/HeGrammarSynthesizer_WithConditionalHyphen.cpp
```diff
@@ -8,6 +8,7 @@
 #include <inflection/util/StringViewUtils.hpp>
 #include <unicode/uchar.h>
 #include <unicode/uscript.h>
+#include <unicode/utf16.h>
```
grhoten, line 11: This type of change seems to be overly cautious, but it's OK to include.

src/inflection/dictionary/metadata/CompressedArray.hpp
```diff
@@ -45,6 +45,11 @@ class inflection::dictionary::metadata::CompressedArray {
     CompressedArray& operator=(const CompressedArray& other) = delete;
 };
 
+namespace inflection::dictionary::metadata {
+template <typename T>
+CompressedArray(const ::std::vector<T>&) -> CompressedArray<T>;
+}
+
```
grhoten: This feels like a false alarm.

Do you really need to support `-Wctad-maybe-unsupported`? The code passes its tests on Linux, macOS, and Windows. It seems to work with g++ and clang++. What OS/compiler combination would fail for this configuration?

If you have a concern about portability to other operating systems like Windows, there is a long list of warnings from the MSVC compiler that could be addressed to improve portability to that operating system.

grhoten (changes requested):

src/inflection/util/ResourceLocator.cpp
nciric: I managed to resolve this issue with internal setting, so this will be reverted.

src/inflection/grammar/synthesis/HeGrammarSynthesizer_WithConditionalHyphen.cpp
nciric: I get hard errors if I don't

error: use of undeclared identifier 'U16_PREV', error: use of undeclared identifier 'U16_GET', error: use of undeclared identifier 'U16_LENGTH'

src/inflection/dictionary/metadata/CompressedArray.hpp
nciric: The -Wctad-maybe-unsupported is on by default and developers in specific environment can't override it, so it's a hard requirement for me.

src/inflection/util/ResourceLocator.cpp
grhoten: I'm also very open to improving the OS portability of ResourceLocator::getDirectoryFromEnvironment for additional operating systems, like Windows. The Windows path management still needs to be figured out.

nciric: That will have to wait for another PR (I am targeting minimal changes in this one, just to make it work internally without patches).

src/inflection/dictionary/metadata/CompressedArray.hpp
grhoten: Instead of this wonkiness, can we just go back to the older C++14 style where it's explicit at the call site? That's how it used to be called before my editor was trying to suggest modern syntax.

If this is a requirement, then `-Wctad-maybe-unsupported` should be added to CMakeLists.txt to avoid the issue from being reintroduced. Perhaps it can be around lines 65-68 with the other compiler warnings. That's around where `-Wglobal-constructors -Wexit-time-destructors` is also used.

grhoten: FYI the `-Wweak-vtables` is defined there too due to a nefarious bug involving multiple vtables defined in a library with full link time optimization turned on. Using unique vtables this way also makes the library slightly smaller.

nciric:
Thanks @grhoten! I've updated the PR per your suggestion:

1. Reverted the deduction guides from the header files (`Finally.hpp`, `CompressedArray.hpp`, `MarisaTrie.hpp`).
2. Made template parameters explicit at all call sites across `tools/` and `test/`.
3. Added `-Wctad-maybe-unsupported` to `CMakeLists.txt` under Clang compile flags to enforce explicit template parameters going forward.

src/inflection/dictionary/metadata/CompressedArray.hpp
nciric: Done

grhoten (approved):

